Issue #40, opened 2020-06-26
Pull in react-axe


WilcoFiers:
As part of the monorepo / scoped package effort, we should migrate react-axe to @axe-core/react. Same way we've done the other packages.

padmavemulapati:
`react` - pulled into axe-core-npm, verified during axe-core-npm testing.

straker:
No docs required

PR #53, opened 2020-07-16
feat: add `@axe-core/react`

michael-siek:
See: #40 

## Reviewer checks


.gitignore
```diff
@@ -1,4 +1,10 @@
 node_modules
 dist
 __tests__
-coverage
+
+packages/react/coverage
+packages/react/example/node_modules
```
WilcoFiers, line 6: This one shouldn't be necessary, as you've got a `node_modules` above.

.gitignore
```diff
@@ -1,4 +1,10 @@
 node_modules
 dist
 __tests__
-coverage
```
WilcoFiers, line 4: Why take this out?

packages/react/.eslintrc.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+  rules: {
+    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/no-unused-vars': 'off',
+    'react/prop-types': 'off',
+    'react/no-find-dom-node': 'off',
+    'no-undef': 'off'
```
WilcoFiers, line 7: I think we should keep these on and disable the rules in the places we need them disabled.

packages/react/.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*]
+indent_style = tab
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
```
WilcoFiers, line 7: From the look of your gitignore, you didn't mean to include this.

packages/react/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Initialize the module
+
+Call the exported function passing in the React and ReactDOM objects as well as a timing delay in milliseconds that will be observed between each component change and the time the analysis starts.
+
+```js
+var React = require('react');
+var ReactDOM = require('react-dom');
+
+if (process.env.NODE_ENV !== 'production') {
+  var axe = require('@axe-core/react');
```
WilcoFiers, line 26:
```suggestion
const React = require('react');
const ReactDOM = require('react-dom');

if (process.env.NODE_ENV !== 'production') {
  const axe = require('@axe-core/react');
```

packages/react/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Shadow DOM
+
+With version 3.0.0, @axe-core/react now runs accessibility tests inside of open Shadow DOM. You don't have to do anything special other than run @axe-core/react on an component encapsulated with open Shadow DOM (as opposed to closed). For more information, see the [axe-core repo](https://github.com/dequelabs/axe-core).
+
+## Configuration
+
+There is a fourth optional argument that is a configuration object for axe-core. Read about the object here: https://github.com/dequelabs/axe-core/blob/master/doc/API.md#api-name-axeconfigure
+
+```js
+var config = {
```
WilcoFiers, line 52:
```suggestion
const config = {
```

packages/react/README.md
```diff
@@ -0,0 +1,114 @@
...
+  ]
+};
+
+axe(React, ReactDOM, 1000, config);
+```
+
+Axe-core's context object can be given as a fifth optional argument to specify which element should (and which should not) be tested. Read more from the Axe-core documentation: https://github.com/dequelabs/axe-core/blob/master/doc/API.md#context-parameter
+
+```js
+var context = {
```
WilcoFiers, line 67:
```suggestion
const context = {
```

packages/react/cypress/integration/index-test.js
```diff
@@ -0,0 +1,65 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import axe from '../../dist/index';
```
WilcoFiers, line 3: Since this is now in a monorepo, we should be able to do this:
```suggestion
import axe from '@axe-core/react';
```

packages/react/example/package.json
```diff
@@ -0,0 +1,45 @@
...
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "prestart": "npm run build",
+    "start": "http-server -u localhost -p 8080",
+    "build": "browserify src/example.js --outfile dist/bundle.js"
+  },
+  "author": "Dylan Barrell (dylan@barrell.com)",
+  "license": "MIT",
+  "dependencies": {
```
WilcoFiers, line 14: And this will need "@axe-core/react"

packages/react/example/src/example.js
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import GlobalHeader from './globalHeader';
+import ServiceChooser from './serviceChooser';
+
+const axe = require('../../dist/index.js');
```
WilcoFiers, line 7: Lets have this import `@axe-core/react`

Issue #68, opened 2020-08-14
axe-core/cli - giving error while executing 


padmavemulapati:
Steps to reproduce:
1. clone latest axe-core-npm develop branch code
2. npm install
3. npm run bootstrap
4. npm lint
5. npm fmt
6. installed cli globally - `npm install @axe-core/cli@next -g`
7.navigated to pakcages\cli
8. tried to run `axe --help` to make sure cli installed properly.  - success
9. Tried to run - `axe https://www.deque.com` -  giving error 
Issue:
`DQ-QA-Team-MBP:cli dq_user$ axe https://www.deque.com
Running axe-core 4.0.1 in chrome-headless

Testing https://www.deque.com ... please wait, this may take a minute.
 
An error occurred while testing this page.
Please report the problem to: https://github.com/dequelabs/axe-cli/issues/ `

padmavemulapati:
Working now, after the fix

PR #69, opened 2020-08-14
fix(cli): start webdriver correctly

michael-siek:
See: #68

test: #72 

## Reviewer checks

### DIFF
--- a/packages/cli/lib/axe-test-urls.js
+++ b/packages/cli/lib/axe-test-urls.js
@@ -4,6 +4,8 @@ const WebDriver = require('selenium-webdriver');
 const AxeBuilder = require('@axe-core/webdriverjs');
 
 async function testPages(urls, config, events) {
+  //selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_ThenableWebDriver.html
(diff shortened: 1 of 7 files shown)


packages/cli/lib/axe-test-urls.js
```diff
@@ -3,15 +3,13 @@
 const WebDriver = require('selenium-webdriver');
 const AxeBuilder = require('@axe-core/webdriverjs');
 
-function testPages(urls, config, events) {
+async function testPages(urls, config, events) {
```
stephenmathieson, line 6: Why is this an `async` function?

michael-siek: I forgot to remove it when testing

michael-siek: https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_ThenableWebDriver.html

straker: Michael and I paired on it. We tried to use the thenable interface as is without `await`ing it, and the unit tests for `startDriver` were causing us so many problems. We spent an hour trying to solve it and just gave up and made it work.

PR #73, opened 2020-08-19
feat: release v4.0.0

michael-siek:
## Reviewer checks


stephenmathieson:
Have we run `npm deprecate` on the old libraries? If not, we should shortly after this lands.

Issue #105, opened 2020-10-19
Support puppeteer v5


stephenmathieson:
Puppeteer v5 has been out for awhile now. The `@axe-core/puppeteer` package has a [`peerDependency` on puppeteer v1/v2](https://github.com/dequelabs/axe-core-npm/blob/v4.0.0/packages/puppeteer/package.json#L47-L49).

Is there a plan to support v5? Will this happen any time soon?

Installing the `@latest` tag of both yields this warning:

```
warning " > @axe-core/puppeteer@4.0.0" has incorrect peer dependency "puppeteer@^1.10.0 || ^2.0.0".
```

stephenmathieson:
```
warning " > @axe-core/puppeteer@4.0.0" has incorrect peer dependency "puppeteer@^1.10.0 || ^2.0.0".
```

michael-siek:
I can take a look at it.

stephenmathieson:
Thanks Michael!

padmavemulapati:
Verified with the latest code `axe-core-npm`, When tried to install puppeteer , it is installing with the version `puppeteer@5.4.0 ` and with this version , able to execute unit tests successfully, as well tested an example test file with simple analyse API. Got the expected results:
![image](https://user-images.githubusercontent.com/47409579/97313055-36465880-188c-11eb-8175-23e9a17d0697.png)
[output_puppeteerV5.txt](https://github.com/dequelabs/axe-core-npm/files/5445862/output_puppeteerV5.txt)

PR #106, opened 2020-10-21
chore(puppeteer): update to support `puppeteer@latest`

michael-siek:
See: #105 

## Reviewer checks

### DIFF
--- a/packages/puppeteer/package.json
+++ b/packages/puppeteer/package.json
@@ -34,7 +34,7 @@
     "express": "^4.16.4",
     "mocha": "^5.2.0",
     "nyc": "^14.1.1",
-    "puppeteer": "latest",
+    "puppeteer": "^5.3.1",
     "sinon": "^7.3.2",
     "source-map-support": "^0.5.9",
     "test-listen": "^1.1.0",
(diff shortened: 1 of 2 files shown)


packages/puppeteer/package.json
```diff
@@ -34,7 +34,7 @@
     "express": "^4.16.4",
     "mocha": "^5.2.0",
     "nyc": "^14.1.1",
-    "puppeteer": "^1.10.0",
+    "puppeteer": "latest",
```
stephenmathieson, line 37: Since there is a `package-lock.json` file here, using `latest` doesn't really mean "latest". I think we ought to pin v5 here instead.

packages/puppeteer/package.json
```diff
@@ -45,7 +45,7 @@
     "axe-core": "^4.0.1"
   },
   "peerDependencies": {
-    "puppeteer": "^1.10.0 || ^2.0.0"
+    "puppeteer": ">=1.10.0"
```
stephenmathieson, line 48: This isn't safe. If Puppeteer v6 removes functionality we rely on, we'll be telling our users that this package works with v6 while it actually does not.

stephenmathieson (changes requested):

packages/puppeteer/package.json
michael-siek: I could do something like 
```json
	"puppeteer": ">=1.10.0 < 6"
```

stephenmathieson: That seems reasonable to me

stephenmathieson (commented):
LGTM! Should get an approval from a maintainer tho 😉

stephenmathieson (approved):

PR #177, opened 2020-12-02
Handle undefined config when not specified

### LINKED ISSUE #176: React: Cannot read property 'runOnly' of undefined
Hello! Just updated to 4.1.0 and noticed new error in my console:
```
Uncaught (in promise) TypeError: Cannot read property 'runOnly' of undefined
    at Module.reactAxe [as default]
```
In the readme, it gives the example of skipping the configuration object to get the context object by passing `undefined`. 
```
// my old config
axe.default(React, ReactDOM, 1000, undefined, { include: [['#app']] });
```

The error disappeared when I passed an empty object instead, but I'm worried that I might be overriding some defaults? If not, I could open a PR to update the docs if you'd like.
```
// new docs change
const context = {
  include: [['#preview']]
};

axe(React, ReactDOM, 1000, {}, context);
```

geoffharcourt:
`react-axe` will fail with an exception if the configuration object is
not specified in the call to `axe()`:

A call like this:

```es6
axe(React, ReactDOM, 3000)
```

Throws an error like this:

```
index.js:374 Uncaught TypeError: Cannot read property 'runOnly' of undefined
    at reactAxe (index.js:374)
```

The existence of the configuration object is checked lower down in the
function, but the early check to see if `runOnly` is relevant doesn't work if
the object isn't there.

A workaround is to specify an empty object `{}`, but it seems from the
type definitions that this is truly meant to be an optional parameter.

Fix #176

## Reviewer checks

### DIFF
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -355,7 +355,7 @@ function reactAxe(
   timeout = _timeout;
   context = _context;
 
-  const runOnly = conf['runOnly'];
+  const runOnly = (conf || {})['runOnly'];
   if (runOnly) {
     conf['rules'] = axeCore
       .getRules(runOnly)
--- a/packages/react/test/index.ts
+++ b/packages/react/test/index.ts
@@ -35,6 +35,9 @@ reactAxe(React, ReactDOM, 1000, {}, context);
 // axe-core context: string
 reactAxe(React, ReactDOM, 1000, {}, '#container');
 
+// axe-core undefined config
+reactAxe(React, ReactDOM, 1000, {}, undefined)
+
 // axe-core context: ContextObject
 reactAxe(
   React,


packages/react/index.ts
```diff
@@ -355,7 +355,7 @@ function reactAxe(
   timeout = _timeout;
   context = _context;
 
-  const runOnly = conf['runOnly'];
+  const runOnly = (conf || {})['runOnly'];
```
geoffharcourt, line 358: We could alternately define a default argument of `{}` for `conf` on line 350, but I wasn't sure if the config call below would change with an empty object vs. skipping the call to `configure` completely.

packages/react/test/index.ts
```diff
@@ -35,6 +35,9 @@ reactAxe(React, ReactDOM, 1000, {}, context);
 // axe-core context: string
 reactAxe(React, ReactDOM, 1000, {}, '#container');
 
+// axe-core undefined config
+reactAxe(React, ReactDOM, 1000, {}, undefined)
```
geoffharcourt, line 39: The call on line 4 doesn't specify an argument, but the suite doesn't catch a failure.

jeeyyy:
@geoffharcourt could you please update your branch with the base branch & push. Thanks.

geoffharcourt:
@jeeyyy done!

jeeyyy:
@geoffharcourt it still says the branch is out of date. Could you please verify?

geoffharcourt:
@jeeyyy someone pushed new code since I rebased.

geoffharcourt:
@jeeyyy rebased and pushed

jeeyyy (changes requested):
@geoffharcourt for some reason the CircleCI workflow for this branch is not triggered. 
It seems like the forked PR does not display in the upstream repo. 

For example here is another [contribution PR](https://github.com/dequelabs/axe-core-npm/pull/181) which seems to have the [pipeline in CircleCI](https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm?branch=pull%2F181)

But for this tree the link is not found - https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm?branch=pull%2F177

Here are some [debug notes](https://discuss.circleci.com/t/fork-pull-request-build-did-not-display-in-upstream-repo/29648)

Could you check/ push a dummy commit to see if Circle CI runs. Thanks.

geoffharcourt:
Hi, this is really growing to be an outsized investment to help fix a bug. We will just work with the workaround instead of fixing the code.

PR #181, opened 2020-12-15
feat(react): Add support for custom logger

guidokessels:
At the moment `@axe-core/react` will always log the results to the console. We have a use case where we want to handle the results in a different way. To be able to do this I added a new optional parameter which can be used to handle the results from `@axe-core/react`. If this parameter is not defined it will fall back to logging to console.

## Reviewer checks


dylanb:
@jeeyyy LGTM but do we want to start adding more tests? If so, a test for this change would be good.

jeeyyy (changes requested):
@guidokessels thanks for the PR ⭐ 
Could you add some tests please?

jeeyyy:
@guidokessels any updates on adding tests to these?

stephenmathieson:
@dylanb IMO it's very unreasonable to ask contributors to add tests. The current test suite is _extremely lacking and very confusing_. Furthermore, it doesn't actually "work" (IIRC you can comment out `axe.run()` and the tests still pass). In order to add good tests, a contributor would first have to create a good test suite. This would be a considerable amount of work, and is not something I believe we can expect from drive-by-contributors.

michael-siek (approved):
Thank you very much @guidokessels for your contribution!

Issue #157, opened 2020-11-03
[@axe-core/react] Uncaught TypeError: Cannot read property 'isConnected' of null


n370:
I'm seeing this error being logged on the console.

https://github.com/dequelabs/axe-core-npm/blob/1f06cb67e2bbbf1922fd9d2e6bcee74f092cfa1c/packages/react/index.ts#L200

I'm wondering which situations `node` is expected to be `null` here.

straker:
Sorry for the super late response. What error is being thrown exactly?

n370:
Hey @straker thanks for getting back on this one. I'd have to find some time to try to reproduce this error again on my end but I think we fixed it by passing an empty object to the `axe` function fourth parameter like:

```
axe(React, ReactDOM, 1000, {});
```
You may be able to reproduce it by omitting that argument or passing `undefined`. I'll try to find some time to reproduce the error again and get back to you here.

PR #188, opened 2021-01-15
fix: `conf` defaults to `{}`

Alex-Joseph:
Description:

version: `4.1.0`

I was getting the error `TypeError: Cannot read property 'runOnly' of undefined` from this line:

`axe.default(React, ReactDom, 1000);`

The solution was similar to [this author's solution](https://github.com/dequelabs/axe-core-npm/issues/157#issuecomment-758724273), therefore perhaps they are related?

## Reviewer checks

### DIFF
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -2,6 +2,7 @@
 import axeCore = require('axe-core');
 import rIC = require('requestidlecallback');
 import after = require('./after');
+import { isEmpty } from 'ramda';
 
 const requestIdleCallback = rIC.request;
 const cancelIdleCallback = rIC.cancel;
@@ -347,7 +348,7 @@ function reactAxe(
   _React: typeof React,
   _ReactDOM: typeof ReactDOM,
   _timeout: number,
-  conf?: ReactSpec,
+  conf = {} as ReactSpec,
   _context?: axeCore.ElementContext
 ): Promise<void> {
   React = _React;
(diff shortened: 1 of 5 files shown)


packages/react/index.ts
```diff
@@ -347,23 +348,23 @@ function reactAxe(
...
+  const runOnly = conf['runOnly'];
   if (runOnly) {
     conf['rules'] = axeCore
       .getRules(runOnly)
       .map(rule => ({ ...rule, id: rule.ruleId, enabled: true }));
     conf['disableOtherRules'] = true;
   }
 
-  if (conf) {
+  if (!isEmpty(conf)) {
```
stephenmathieson, line 367: I don't this this justifies bringing in a dependency. Can we do this without `ramda`?

Alex-Joseph, line 367: Hmm that is a fair point. What about this?

```suggestion
  if (Object.keys(conf).length > 0) {
```

packages/react/index.ts
```diff
@@ -2,6 +2,7 @@
 import axeCore = require('axe-core');
 import rIC = require('requestidlecallback');
 import after = require('./after');
+import { isEmpty } from 'ramda';
```
Alex-Joseph, line 5:
```suggestion

```

Alex-Joseph (commented):
Suggested change

packages/react/index.ts
stephenmathieson: Looks good to me!

n370:
👍

jeeyyy (approved):

stephenmathieson:
Thanks @Alex-Joseph!

Issue #75, opened 2020-08-21
axe-webdriverio does not work


straker:
Axe-webdriverio hasn't been updated in 3 years and is out-of-date with latest webriverio changes. 

Here's the problem: axe-webdriverio is a wrapper around axe-webdriverjs and does that by bridging the webdriverio functions to webdriverjs functionality (https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverio/lib/bindings.js). However, the functions it tries to wrap are no longer available in webdriverio (element and frame do not exist anymore). This causes webdriverjs to fail when injecting axe into iframes as those functions do not exist on the driver (https://github.com/dequelabs/axe-webdriverjs/blob/develop/lib/axe-injector.js#L62-L91)

Theoretically the fix should be to wrap the correct functions (`findElement` and `switchToFrame`), but we should add proper integration tests to make sure this truly works as right now there is only a single unit test. Also, the unit test is bad because it uses a mock driver, meaning updating to v6 of webdriverio (which should fail) is passing the tests still.

If you test with an actual browser instance you'll see it failing:

```js
const bindings = require('../lib/bindings');
const { remote } = require('webdriverio');

let browser;
before(async () => {
  browser = await remote({
    logLevel: 'trace',
    capabilities: {
        browserName: 'chrome'
    }
  });
});
```

jeeyyy:
Related issue - https://github.com/dequelabs/axe-core-npm/issues/58

padmavemulapati:
Tried to execute a testscript - with simple analyse() api, its resulting with error: saying 
```
MacBook-Pro:webdriverio dq_user$ node test.js
2021-03-03T11:29:34.761Z INFO devtools:puppeteer: Initiate new session using the DevTools protocol
[...]

Issue #58, opened 2020-08-05
add a readme to webdriverio


WilcoFiers:
The readme seems pretty incomplete, we should get it a similar readme to what we have for webdriverjs or puppeteer.

padmavemulapati:
Verified the ReadME file

michael-siek:
Closing: README verified

PR #196, opened 2021-02-17
refactor(webdriverio): rewrite webdriverio to TS

michael-siek:
See: #58, #75 

## Reviewer checks


packages/webdriverio/.eslintrc.js
```diff
@@ -1,15 +1,7 @@
 module.exports = {
   rules: {
-    '@typescript-eslint/no-empty-function': 'off',
+    '@typescript-eslint/no-explicit-any': 'off',
```
stephenmathieson, line 3: I don't think we should disable this rule. Using `any` defeats the purpose of TypeScript. Can we change this to a warning instead?

```suggestion
    '@typescript-eslint/no-explicit-any': 'warn',
```

stephenmathieson (commented):
LGTM! Should get someone on your team to give the final approval tho 👍

packages/webdriverio/.eslintrc.js
michael-siek: I could just remove this from eslint and add `// @ts-ignore` where I explicitly use `any` type which is only in test files.

stephenmathieson: If it's only in test files, then I'm strongly against turning this rule off globally. Just use an `override` instead:

```json
{
  "overrides": [
    {
      "files": ["*.test.ts", "*.test.tsx"],
      "rules": {
        "@typescript-eslint/no-explicit-any": "off"
      }
    }
  ]
}
```

stephenmathieson: I usually have something like this in my `eslintrc`:

```js
{
  "overrides": [
    {
      "files": ["*.test.ts", "*.test.tsx"],
      "rules": {
        // Tests can use `any`.
        "@typescript-eslint/no-explicit-any": "off",
        // Noop mocks, etc.
        "@typescript-eslint/no-empty-function": "off",
        // Tests can disable the TS compiler.
        "@typescript-eslint/ban-ts-comment": "off"
      }
    }
  ]
}
```

packages/webdriverio/README.md
```diff
@@ -1,5 +1,140 @@
...
+  try {
+    const results = await builder.analyze();
+    console.log(results);
+  } catch (e) {
+    console.error(e);
+  }
+})();
+```
+
+## AxeBuilder({ client: WebdriverIO.BrowserObject })
```
jeeyyy, line 54: Nitpick, all of these API methods should be nested under a level 2 heading API and the rest should be level 3 in my opinion.

packages/webdriverio/README.md
```diff
@@ -1,5 +1,140 @@
...
+```js
+new AxeBuilder({ client }).include('.results-panel');
+```
+
+## AxeBuilder#exclude(selector: String)
+
+Add a CSS selector to the list of elements to exclude from analysis
+
+```js
+new AxeBuilder({ client }).include('.some-element').exclude('.another-element');
```
jeeyyy, line 75:
```suggestion
new AxeBuilder({ client }).exclude('.another-element');
```

packages/webdriverio/README.md
```diff
@@ -1,5 +1,140 @@
...
+```
+
+## AxeBuilder#options(options: [axe.RunOptions](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#options-parameter))
+
+Specifies options to be used by `axe.a11yCheck`. Will override any other configured options. including calls to `AxeBuilder#withRules()` and `AxeBuilder#withTags()`. See [axe-core API documentation](https://github.com/dequelabs/axe-core/blob/master/doc/API.md) for information on its structure.
```
jeeyyy, line 80: Out of curiosity What is `axe.a11yCheck`?

packages/webdriverio/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  private client: BrowserObject;
+  private axeSource: string;
+  private includes: string[];
+  private excludes: string[];
+  private option: RunOptions;
+  private disableFrameSelectors: string[];
+  constructor({ client }: Options) {
+    assert(
+      isWebdriverClient(client),
+      'An instantiated webdriverio client of v5 or v6 is required.'
```
jeeyyy, line 28: can we say this - https://github.com/dequelabs/axe-devtools-npm/blob/develop/packages/webdriverio/src/index.ts#L107

Given newer version 7, is coming out soon?

packages/webdriverio/src/utils/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  DoneFunction
+} from '../types';
+
+/**
+ * Validates that the client provided is WebdriverIO v5 or v6.
+ * @param {any} client
+ * @returns {BrowserObject}
+ */
+
+export const isWebdriverClient = (client: any): client is BrowserObject => {
```
jeeyyy, line 15:
```suggestion
export const isWebdriverClient = (client: BrowserObject): boolean => {
```

packages/webdriverio/src/utils/index.ts
```diff
@@ -0,0 +1,105 @@
...
+ * @param {AnalyzePageParams} analyzeContext
+ * @param {DoneFunction} done
+ * @returns {AnalyzePageResponse}
+ */
+
+export const analyzePage = (
+  analyzeContext: AnalyzePageParams,
+  done: DoneFunction
+): AnalyzePageResponse => {
+  const axeCore = (window as any).axe;
```
jeeyyy, line 72: You should do this - https://github.com/dequelabs/axe-devtools-npm/blob/develop/packages/cypress/src/types.ts#L15

packages/webdriverio/src/utils/index.ts
```diff
@@ -0,0 +1,105 @@
...
+
+/**
+ * Checks to make sure that the error thrown was not a stale iframe
+ * @param {Error} error
+ * @returns {void}
+ */
+
+export const logOrRethrowError = (error: Error): void => {
+  if (
+    (error as any).seleniumStack?.type === 'StaleElementReference' ||
```
jeeyyy, line 96: nitpick, you could extend the Error interface same as global window instead of casting as any

jeeyyy (changes requested):
Some feedback inline.

packages/webdriverio/README.md
stephenmathieson: It's the _old_ axe API. This was probably originally written years ago 🤷‍♂️

packages/webdriverio/src/index.ts
```diff
@@ -0,0 +1,230 @@
...
+  Element
+} from './types';
+import {
+  analyzePage,
+  isWebdriverClient,
+  normalizeContext,
+  logOrRethrowError
+} from './utils';
+
+export default class AxeBuilder {
```
stephenmathieson, line 18: Is it "safe" to change the export style here? The previous version used CommonJS-style exports I believe.

jeeyyy (approved):

jeeyyy (approved):

Issue #195, opened 2021-02-09
[@axe-core/webdriverjs] Typescript issue, missing type declaration file


J-VU:
Type file _@types/axe-webdriverjs_ does not work with current webdriverjs

**Reproduce**
Create a typescript project using @axe-core/webdriverjs

**index.ts**
```
import AxeBuilder from '@axe-core/webdriverjs'
import WebDriver from 'selenium-webdriver'

const driver = new WebDriver.Builder().forBrowser('firefox').build()
driver.get('https://dequeuniversity.com/demo/mars/').then(function () {
  new AxeBuilder(driver).analyze(function (err, results) {
    if (err) {
      // Handle error somehow
    }
    console.log(results)
  })
})
```

**tsconfig.json**
```
{
  "compilerOptions": {
    "lib": ["es2015", "es2017", "dom", "es2020"],
    "module": "commonjs",
    "moduleResolution": "node",
    "target": "es2017",
    "noImplicitReturns": true,
    "noFallthroughCasesInSwitch": true,
    "pretty": true,
    "strict": true,
    "sourceMap": true,
    "outDir": "./dist",
    "declaration": false,
    "skipLibCheck": true,
    "noImplicitAny": true,
    "esModuleInterop": true,
    "resolveJsonModule": true,
    "baseUrl": "./",
    "paths": {
      "*": ["src/@types/*"]
    }
  },
  "include": ["src/**/*.ts"]
}
```

michael-siek:
I tested this out and to add more context to this ticket it looks like `@types/axe-webdriverjs` doesn't support our 'new' scoped package `@axe-core/webdriverjs` so it produces this error
```console
implicitly has an 'any' type.
  Try `npm i --save-dev @types/axe-core__webdriverjs` if it exists or add a new declaration (.d.ts) file containing `declare module '@axe-core/webdriverjs';`ts(7016)
```

pauchan:
Hi @michael-siek , thanks for the fix!
I'm hitting the same issue when trying to import `@types/axe-webdriverjs` in one of my projects. 
Would it be possible for you to consider releasing new version of the package with this fix in place?

michael-siek:
[...]

PR #197, opened 2021-02-19
refactor(webdriverjs): rewrite webdriverjs to TS

michael-siek:
See: #195 

## Reviewer checks


packages/webdriverjs/.eslintrc.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  rules: {
+    '@typescript-eslint/no-var-requires': 'off'
```
stephenmathieson, line 3: Why are you disabling this rule?

packages/webdriverjs/README.md
```diff
@@ -36,140 +36,115 @@ driver.get('https://dequeuniversity.com/demo/mars/').then(function () {
...
     // Handle error somehow
   }
-  console.log(results);
-});
+  console.log(results)
+})
 ```
 
-### AxeBuilder#analyze(callback:Function)
+### AxeBuilder#analyze(): Promise<axe.Results | Error>
```
stephenmathieson, line 128: This reads as:

The class `AxeBuilder`'s method `analyze` returns a `Promise` which resolves to `axe.Results` or an `Error`, which is not correct. The `Promise` will never _resolve_ with an error, but it may `reject` instead.

IMO this should be:

```suggestion
### AxeBuilder#analyze(): Promise<axe.Results>
```

packages/webdriverjs/example.ts
```diff
@@ -0,0 +1,14 @@
+import AxeBuilder from './src';
+import * as WebDriver from 'selenium-webdriver';
+
+const driver = new WebDriver.Builder().forBrowser('chrome').build();
+
+driver.get('https://html5-sandbox.glitch.me/').then(() => {
+  const axe = new AxeBuilder({ driver, builderOptions: { noSandbox: true } });
+  axe.analyze((err, results) => {
```
stephenmathieson, line 8: Thoughts on calling this file `example-with-callback.js` and creating an `example.js` which uses `async/await` to get results? At this point, I think most developers are familiar enough with `async/await` for the syntax to be used in the "main example".

packages/webdriverjs/package.json
```diff
@@ -1,37 +1,19 @@
 {
   "name": "@axe-core/webdriverjs",
   "version": "4.1.1",
-  "description": "Provides a method to inject and analyze web pages using aXe",
+  "description": "Provides a method to inject and analyze web pages using axe",
+  "author": "Michael Siek<me@michaelsiek.com>",
```
stephenmathieson, line 5: I think it's fine to add yourself to `contributors`, but claiming ownership of this package probably won't fly...

packages/webdriverjs/package.json
```diff
@@ -1,37 +1,19 @@
...
-      "url": "http://github.com/wilcofiers/"
-    },
-    {
-      "name": "David Sturley",
-      "organization": "Deque Systems, Inc.",
-      "url": "http://deque.com/"
-    }
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist/index.d.ts"
```
stephenmathieson, line 10: Is this the only file we need in the tarball?

packages/cli/lib/axe-test-urls.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 const WebDriver = require('selenium-webdriver');
-const AxeBuilder = require('@axe-core/webdriverjs');
+const AxeBuilder = require('@axe-core/webdriverjs').default;
```
stephenmathieson, line 4: This indicates a breaking change. We can't ship this until we're ready for a SEMVER major release.

If we want this to land sooner than that, we'll want to do some hackery the TS:

```
export default AxeBuilder
```

should be:

```
export = AxeBuilder
```

Alternatly, we could add a `src/index.ts` which looks like:

```
import AxeBuilder from './AxeBuilder'

// Enable ESM/TS imports while we wait for a SEMVER major release.
AxeBuilder.default = AxeBuilder

export = AxeBuilder
```

stephenmathieson (changes requested):
I don't think we can make breaking changes here right now.

packages/cli/lib/axe-test-urls.js
```diff
@@ -57,7 +57,7 @@ async function testPages(urls, config, events) {
       })
       .then(() => {
         // Set everything up
-        const axe = new AxeBuilder(driver, config.axeSource);
+        const axe = new AxeBuilder({ driver, axeSource: config.axeSource });
```
stephenmathieson, line 60: This also indicates a breaking change. We can't do this either.

packages/webdriverjs/README.md
```diff
@@ -36,140 +36,115 @@ driver.get('https://dequeuniversity.com/demo/mars/').then(function () {
...
 ```
 
-If you wish to run a specific version of axe-core, you can pass the source axe-core source file in as a string. Doing so will mean axe-webdriverjs runs this version of axe-core, instead of the one installed as a dependency of axe-webdriverjs.
+If you wish to run a specific version of [axe-core](https://github.com/dequelabs/axe-core), you can pass the source of axe-core source file in as a string. Doing so will mean `@axe-core/webdriverjs` run this version of axe-core, instead of the one installed as a dependency of `@axe-core/webdriverjs`.
 
-```javascript
-var axeSource = fs.readFileSync('./axe-1.0.js', 'utf8');
-var builder = new AxeBuilder(driver, axeSource);
+```js
+const axeSource = fs.readFileSync('./axe-1.0.js', { encoding: 'utf-8' });
```
stephenmathieson, line 50: Why specify the encoding like this? Passing a string is _much_ more common here.

```suggestion
const axeSource = fs.readFileSync('./axe-1.0.js', 'utf8');
```

packages/webdriverjs/example-with-callback.ts
```diff
@@ -1,10 +1,11 @@
-const AxeBuilder = require('./lib');
-const WebDriver = require('selenium-webdriver');
+import AxeBuilder from './src';
```
stephenmathieson, line 1: IMO we should leave the examples as regular/vanilla JavaScript. This allows them to be copy/pasted and executed without needing a build step.

packages/webdriverjs/package.json
```diff
@@ -1,37 +1,21 @@
...
-    {
-      "name": "Wilco Fiers",
-      "organization": "Deque Systems, Inc.",
-      "url": "http://github.com/wilcofiers/"
-    },
-    {
-      "name": "David Sturley",
-      "organization": "Deque Systems, Inc.",
-      "url": "http://deque.com/"
+      "name": "Michael Siek (me@michaelsiek.com)"
```
stephenmathieson, line 7: Should leave the other contributors here.

packages/webdriverjs/src/index.ts
```diff
@@ -0,0 +1,199 @@
+import type { WebDriver } from 'selenium-webdriver';
+import type { RunOptions, Spec, AxeResults } from 'axe-core';
+import { source } from 'axe-core';
+import type { CallbackFunction, BuilderOptions } from './types';
+import { normalizeContext } from './utils';
+import AxeInjector from './axe-injector';
+
+export default class AxeBuilder {
```
stephenmathieson, line 8: Why are you exporting this here and again at the end of the file?

stephenmathieson (commented):
Small stuff. Looking good tho 👍

stephenmathieson (commented):
LGTM 👍 

You should get an OK from @jeeyyy, @WilcoFiers or @AdnoC before merging tho.

jeeyyy (approved):

Issue #35, opened 2020-06-11
rewrite `@axe-core/cli` to ts


michael-siek:

jeeyyy:
WIP - https://github.com/dequelabs/axe-core-npm/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc

padmavemulapati:
Pulled the latest code for axe-core-npm and verified, its now running "**/**.test.ts" file instead "*.js" on cli

PR #198, opened 2021-02-22
refactor(cli): rewrite CLI to TS

michael-siek:
Blocked by: #197 
See: #35 

## Reviewer checks


packages/cli/axe-cli
```diff
@@ -1,3 +1,3 @@
 #!/usr/bin/env node
 
-require('./index')
\ No newline at end of file
+require('./dist/src/bin/cli.js')
```
stephenmathieson, line 3: I've never understood the purpose of this file. Why not just point `package.json#bin` to the `cli.js` file?

packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -1,42 +1,44 @@
-'use strict';
+import * as WebDriver from 'selenium-webdriver';
+// @ts-ignore
```
stephenmathieson, line 2: Why?

michael-siek: Have to wait for this PR to land to remove the `@ts-ignore` https://github.com/dequelabs/axe-core-npm/pull/197

packages/cli/src/bin/index.ts
```diff
@@ -0,0 +1,138 @@
...
+      process.stdout.write(JSON.stringify(outcome, null, 2));
+      return;
+    }
+
+    if (timer) {
+      console.timeEnd('Total test time');
+    }
+
+    /* istanbul ignore if */
+    if (Array.isArray(outcome) && outcome.length > 1) {
```
jeeyyy, line 94: For my context, why do we check > 1 here? I am not aware of what the outcome array holds.

packages/cli/src/types.ts
```diff
@@ -0,0 +1,17 @@
+import type { AxeResults } from 'axe-core';
+
+export interface EventParams {
+  silentMode: boolean;
+  timer: boolean;
+  cliReporter: (...args: any[]) => void;
```
jeeyyy, line 6: we should type this, it is our own method...

packages/cli/src/lib/utils.ts
```diff
@@ -0,0 +1,130 @@
+import * as fs from 'fs';
+import * as path from 'path';
+import * as colors from 'colors';
+import { AxeResults } from 'axe-core';
+
+export const saveOutcome = (
+  outcome: AxeResults | AxeResults[],
+  fileName?: string,
+  dir?: string
+): string | void => {
```
jeeyyy, line 10:
```suggestion
): string => {
```

packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -1,42 +1,43 @@
...
+import * as WebDriver from 'selenium-webdriver';
+import AxeBuilder from '@axe-core/webdriverjs';
+import { AxeResults } from 'axe-core';
+import { EventResponse } from '../types';
 
-const WebDriver = require('selenium-webdriver');
-const AxeBuilder = require('@axe-core/webdriverjs');
+const testPages = async (
+  urls: string | string[],
+  config: any,
```
jeeyyy, line 8: why can't we type this?

packages/cli/src/lib/axe-test-urls.test.ts
```diff
@@ -67,25 +67,25 @@ describe('testPages', function () {
...
-    const asyncScripts = [];
-    mockDriver.executeAsyncScript = async script => {
+    const asyncScripts: string[] = [];
+    mockDriver.executeAsyncScript = async (script: string) => {
       asyncScripts.push(script);
       return script;
     };
 
-    await testPages(['http://foo'], config, {});
+    await testPages(['http://foo'], config, {} as any);
```
jeeyyy, line 88: looks like the third argument can be optional, lets make it `events?: EventResponse` instead of passing an empty object.

packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -1,42 +1,43 @@
...
+  events: EventResponse
```
jeeyyy, line 9:
```suggestion
  events?: EventResponse = {}
```

packages/cli/src/lib/axe-test-urls.test.ts
```diff
@@ -1,60 +1,60 @@
...
+  let config: any;
+  let mockDriver: any;
 
   beforeEach(() => {
     mockDriver = {
-      get: async arg => arg,
-      executeAsyncScript: async arg => arg,
-      executeScript: async arg => arg,
-      wait: async arg => arg,
+      get: async (arg: any) => arg,
```
jeeyyy, line 12: extract `async (arg: any) => arg` to be a variable & simply assign that to each prop here.

packages/cli/src/lib/axe-test-urls.test.ts
```diff
@@ -1,60 +1,60 @@
...
-const assert = require('chai').assert;
-const testPages = require('../lib/axe-test-urls');
+import 'mocha';
+import { assert } from 'chai';
+import testPages from './axe-test-urls';
 
 describe('testPages', function () {
   this.timeout(10000);
-  let config, mockDriver;
+  let config: any;
```
jeeyyy, line 7: Once you type `config` this can go away.

jeeyyy (dismissed):
See inline comments

packages/cli/src/types.ts
michael-siek: I would normally agree but this method is similar to a `console.log` so I think in this instance a type of `any[]` would be good for this

michael-siek: This method is used similarly to a `console.log()` and I think type `any[]` is ok in this instance.

jeeyyy (approved):

PR #209, opened 2021-03-14
fix(webdriverjs): Recursively find <frame> 

yokra9:
Fixed to recursively find `<frame>`  and inject a script into them, not only `<iframe>`.

`<frame>` is deprecated in HTML5, but sometimes use in bad examples when teach modern & accessible coding.
ex: https://barrierfree.nict.go.jp/accessibility/whatsacs/sample/level0/index.html

Thanks!

## Reviewer checks


michael-siek:
Hi, @yokra9 do you mind providing tests for this fix.

yokra9:
@michael-siek thanks for comment! added 2 commits for providing tests for this fix.

* added unit test for this fix.
  + revive test fixtures, `inner-configure-frame.html`, which might have removed unintentionally.
* removed unused test fixtures.

WilcoFiers:
@yokra9 I just wanted to let you know we're still looking to see if we want this change added. As you pointed out, frames were deprecated years ago. Axe has never supported them. We're currently working to improve consistency between the various axe packages we maintain. That includes whether or not they support frames. If we add frame support to webdriverjs, we need to do it to webdriverio, puppeteer, cypress, Java/Ruby/C# selenium, Ruby Watir, etc. That is a fairly substantial amount of work.

I very much do appreciate you setting up this PR. And I apologies for not noticing this before you made additional commits to it. I'm going to bring this up internally and will get back to you soon.

michael-siek:
Thank you very much for your contribution @yokra9.

michael-siek (approved):

PR #213, opened 2021-03-19
chore(webdriverio): fix `switchToParentFrame()`

michael-siek:
See: https://github.com/dequelabs/axe-core-npm/issues/75

## Reviewer checks

### DIFF
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -225,6 +225,10 @@ export default class AxeBuilder {
   private async setBrowsingContext(
     id: null | Element | BrowserObject = null
   ): Promise<void> {
-    await this.client.switchToFrame(id);
+    if (id) {
+      await this.client.switchToFrame(id);
+    } else {
+      await this.client.switchToParentFrame();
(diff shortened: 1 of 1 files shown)


straker (approved):
Even though the docs for `switchToFrame` allow `null` for `id`, it appears they don't handle a case when the page has no frames to switch to.

We can see that it only handles `null` when there is a parentFrame, so the bug is when it's `null` and there is no parent frame it tries to get a frame by the id as if it were not `null`

https://github.com/webdriverio/webdriverio/blob/v6.12.1/packages/devtools/src/commands/switchToFrame.ts#L21-L26

However, in `switchToParentFrame` they actually handle no parent correctly and just return early.

https://github.com/webdriverio/webdriverio/blob/v6.12.1/packages/devtools/src/commands/switchToParentFrame.ts#L19-L21

Since I can't figure how to get access to that `getPageHandle` (it's `undefined` off of the `client`) so we could check before even calling `switchToFrame`, this is probably the best we can do, though it could break if `switchToParentFrame` stops `null` checking properly.

Issue #229, opened 2021-04-22
CLI: unable to execute `@axe-core/cli`


michael-siek:
When using `@axe-core/cli@next` I am getting an error where axe is unable to find `axe.js`
```console
(node:41199) UnhandledPromiseRejectionWarning: Error: ENOENT: no such file or directory, open '/usr/local/lib/node_modules/@axe-core/cli/dist/src/lib/node_modules/axe-core/axe.js'
    at Object.openSync (fs.js:476:3)
    at Object.readFileSync (fs.js:377:35)
    at Object.getAxeSource (/usr/local/lib/node_modules/@axe-core/cli/dist/src/lib/utils.js:82:15)
    at /usr/local/lib/node_modules/@axe-core/cli/dist/src/bin/index.js:20:30
    at Generator.next (<anonymous>)
    at /usr/local/lib/node_modules/@axe-core/cli/dist/src/bin/index.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/usr/local/lib/node_modules/@axe-core/cli/dist/src/bin/index.js:4:12)
    at Command.cli (/usr/local/lib/node_modules/@axe-core/cli/dist/src/bin/index.js:17:28)
    at Command.listener [as _actionHandler] (/usr/local/lib/node_modules/@axe-core/cli/node_modules/commander/index.js:922:31)
(Use `node --trace-warnings ...` to show where the warning was created)
(node:41199) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:41199) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

padmavemulapati:
Verified as part of axe-core-npm QA, not able to reproduce. Now its working fine

PR #232, opened 2021-04-26
fix(cli): find axe-core when installed globally

michael-siek:
See: #229 
This issue only happens when installing `@axe-core/cli` globally.

## Reviewer checks


stephenmathieson (changes requested):
I think we could better solve this using Node's module lookup algorithm.

```
axe_core_file_path = require.resolve('axe-core')
```

This will return the _nearest_ installed axe-core. It works both when installed globally and as a project dependency.

WilcoFiers (dismissed):
Tests

stephenmathieson:
Writing a proper test for this will be difficult. To actually reproduce the problem, we'd have to globally install the program and run it. Not really something that qualities as a "unit test" IMO. Furthermore, we're using Node's algorithm now, which we already assume works (otherwise, nothing here would). Since the existing tests pass, and this is _removing_ custom logic and replacing it with Node's `require`, do we really need to setup an E2E test for this?

packages/cli/src/lib/utils.test.ts
```diff
@@ -71,6 +72,24 @@ describe('utils', () => {
...
+          }
+        });
+      });
+
+      afterEach(() => {
+        mock.restore();
+      });
+      it('fall back to use `locally` installed axe-core', () => {
+        const axeSource = utils.getAxeSource();
+        assert.isNotNull(axeSource);
```
WilcoFiers, line 90: I'm not sure what this is supposed to prove, but just asserting that it isn't null doesn't mean much.

WilcoFiers (changes requested):

WilcoFiers (approved):

Issue #233, opened 2021-04-27
`@axe-core/*` v4.2 release 


michael-siek:
With [axe-core@4.2](https://github.com/dequelabs/axe-core/releases/tag/v4.2.0) release subsequent releases need to happen to all `axe-core-*` integrations. 

- [x] `@axe-core/cli`
- [x] `@axe-core/puppeteer`
- [x] `@axe-core/react`
- [x] `@axe-core/reporter-earl`
- [x] `@axe-core/webdriverio`
- [x] `@axe-core/webdriverjs`

padmavemulapati:
QAed axe-core-npm4.2.0, and shared the report for the same across the team,
https://dequeqa.testrail.io/index.php?/reports/view/1113 - can find the report here

michael-siek:
Closing this is complete

PR #240, opened 2021-05-03
feat: update axe-core to 4.2.0

michael-siek:
See: #233

### DIFF
--- a/packages/puppeteer/src/browser.ts
+++ b/packages/puppeteer/src/browser.ts
@@ -30,8 +30,6 @@ export function configureAxe(config?: Axe.Spec) {
 
   window.axe.configure({
     allowedOrigins: ['<unsafe_all_origins>'],
-    branding: {
-      application: 'axe-puppeteer'
-    }
+    branding: { application: 'axe-puppeteer' }
   });
 }
--- a/packages/puppeteer/src/browser.ts
+++ b/packages/puppeteer/src/browser.ts
@@ -33,5 +33,8 @@ export function configureAxe(config?: Axe.Spec) {
       application: 'axe-puppeteer'
     }
   };
-  window.axe.configure(brandingConfig);
+  window.axe.configure({
+    allowedOrigins: ['<unsafe_all_origins>'],
+    ...brandingConfig
+  });
 }
--- a/packages/puppeteer/src/browser.ts
+++ b/packages/puppeteer/src/browser.ts
@@ -28,13 +28,10 @@ export function configureAxe(config?: Axe.Spec) {
     window.axe.configure(config);
   }
 
-  const brandingConfig = {
+  window.axe.configure({
+    allowedOrigins: ['<unsafe_all_origins>'],
     branding: {
       application: 'axe-puppeteer'
     }
-  };
-  window.axe.configure({
-    allowedOrigins: ['<unsafe_all_origins>'],
-    ...brandingConfig
   });
 }
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -179,8 +179,10 @@ export default class AxeBuilder {
   private get script(): string {
     return `
       ${this.axeSource}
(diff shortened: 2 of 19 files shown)


packages/webdriverio/src/index.ts
```diff
@@ -179,6 +179,7 @@ export default class AxeBuilder {
   private get script(): string {
     return `
       ${this.axeSource}
+      axe.configure({ allowedOrigins: ['<unsafe_all_origins>'] })
```
stephenmathieson, line 182: Any reason not to merge this with the `axe.configure()` call below?

packages/webdriverjs/src/axe-injector.ts
```diff
@@ -75,6 +75,7 @@ export default class AxeInjector {
     return `
     ${this.axeSource}
     ${this.config ? `axe.configure(${this.config})` : ''}
+    axe.configure({ allowedOrigins: ['<unsafe_all_origins>'] })
```
stephenmathieson, line 78: Same merging question.

stephenmathieson (approved):

packages/webdriverio/src/index.ts
michael-siek: We set it set `axe.configure()` so that we can inject axe and get the results from all iframes https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#allowedorigins instead of just same origin iframes

stephenmathieson: Yes.

My question is why we're not doing this:

```
axe.configure({
  allowedOrigins: [...],
  branding: { application: 'webdriverio' }
})
```

NBD tho, which is why I approved.

stephenmathieson (approved):
LGTM

I think long-term, we should make the `allowedOrigins` thing configurable.

packages/puppeteer/src/browser.ts
```diff
@@ -33,5 +33,8 @@ export function configureAxe(config?: Axe.Spec) {
       application: 'axe-puppeteer'
     }
   };
-  window.axe.configure(brandingConfig);
+  window.axe.configure({
+    allowedOrigins: ['<unsafe_all_origins>'],
+    ...brandingConfig
+  });
```
stephenmathieson, line 39: What is the purpose of having `brandingConfig` hoisted up? Seems like this would be just fine:

```js
window.axe.configure({
  allowedOrigins: ['<unsafe_all_origins>'],
  branding: { application: 'axe-puppeteer' }
})
```

Another not a big deal/not worth changing thing btw 🤷

packages/webdriverio/src/index.ts
```diff
@@ -179,7 +179,10 @@ export default class AxeBuilder {
   private get script(): string {
     return `
       ${this.axeSource}
-      axe.configure({ branding: { application: 'webdriverio' }})
+      axe.configure({ 
+        allowedOrigins: ['<unsafe_all_origins>'],
+        branding: { application: 'webdriverio'
```
stephenmathieson, line 184:
```suggestion
        branding: { application: 'webdriverio' }
```

packages/webdriverjs/src/axe-injector.ts
```diff
@@ -75,7 +75,10 @@ export default class AxeInjector {
     return `
     ${this.axeSource}
     ${this.config ? `axe.configure(${this.config})` : ''}
-    axe.configure({ branding: { application: 'webdriverjs' }})
+    axe.configure({ 
+      allowedOrigins: ['<unsafe_all_origins>'], 
+      branding: { application: 'webdriverjs'
```
stephenmathieson, line 80:
```suggestion
      branding: { application: 'webdriverjs' }
```

stephenmathieson (changes requested):
Missed `}` in a couple spots.

straker (approved):

stephenmathieson (approved):

Issue #241, opened 2021-05-04
support puppeteer v9


michael-siek:
puppeteer is currently at v9 and the peerdependency for `@axe-core/puppeteer` is at v6 we should support the latest version of puppeteer.

padmavemulapati:
When simply installed puppeteer - its installing puppeteer@5.5.0 , but installed puppeteer@latest - it installed
```PS C:\Users\DQ_Padmavathi\Desktop\npmcore\axe-core-npm\packages\puppeteer> npm install puppeteer@latest

> puppeteer@9.1.1 install C:\Users\DQ_Padmavathi\Desktop\npmcore\axe-core-npm\packages\puppeteer\node_modules\puppeteer
> node install.js

Downloading Chromium r869685 - 155.9 Mb [====================] 100% 0.0s
```
and example file executed and got the expected results with puppeteer@9.1.1

michael-siek:
Closing no docs required.

PR #242, opened 2021-05-04
feat(puppeteer): support puppeteer v9

michael-siek:
See: #241

### DIFF
--- a/packages/puppeteer/package.json
+++ b/packages/puppeteer/package.json
@@ -34,7 +34,7 @@
     "express": "^4.16.4",
     "mocha": "^5.2.0",
     "nyc": "^14.1.1",
-    "puppeteer": "^5.3.1",
+    "puppeteer": "9.1.0",
     "sinon": "^7.3.2",
     "source-map-support": "^0.5.9",
     "test-listen": "^1.1.0",
@@ -45,7 +45,7 @@
     "axe-core": "^4.2.0"
   },
   "peerDependencies": {
-    "puppeteer": ">=1.10.0 < 6"
+    "puppeteer": ">=1.10.0 < 9.1.0"
   },
   "engines": {
     "node": ">=6.4.0"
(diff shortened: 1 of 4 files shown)


packages/puppeteer/src/axePuppeteer.ts
```diff
@@ -17,7 +17,7 @@ interface IInjectAxeArgs {
   args?: any[];
 }
 
-function injectJSModule(frame: Frame): Promise<void> {
+function injectJSModule(frame: Frame): Promise<ElementHandle<Element> | void> {
```
stephenmathieson, line 20: Might want to do the same hack as below to avoid this weird/useless `return` type:

```js
return frame.addScriptTag({ /* ... */ )}.then(() => undefined)
```

packages/puppeteer/package.json
```diff
@@ -34,7 +34,7 @@
     "express": "^4.16.4",
     "mocha": "^5.2.0",
     "nyc": "^14.1.1",
-    "puppeteer": "^5.3.1",
+    "puppeteer": "9.1.0",
```
stephenmathieson, line 37: Do we want this pinned at `9.1.0`?

```suggestion
    "puppeteer": "^9",
```

packages/puppeteer/package.json
```diff
@@ -45,7 +45,7 @@
     "axe-core": "^4.2.0"
   },
   "peerDependencies": {
-    "puppeteer": ">=1.10.0 < 6"
+    "puppeteer": ">=1.10.0 < 9.1.0"
```
stephenmathieson, line 48: Why `9.1.0`? What makes you think this library won't work with v9.1.1?

stephenmathieson, line 48:
```suggestion
    "puppeteer": ">=1.10.0 < 9"
```

stephenmathieson (changes requested):
The peer dep version should not be so strict.

stephenmathieson (approved):

PR #243, opened 2021-05-05
chore: use public npm registry

michael-siek:


straker (approved):
We should add a tech debt ticket to figure out why circle doesn't fail a build when it's changed to use agora.

Issue #100, opened 2020-10-08
playwright: intergration


michael-siek:
Based on the original issue https://github.com/dequelabs/axe-puppeteer/issues/70 it looks like this would be a nice feature to have.

stephenmathieson:
I am building E2E tests for the new axe app and am using playwright. I've had to hack my own integration together for this. It'd be great if we had an officially supported integration instead!

leeuwe:
I'd like to upvote this. I'd just LOVE to move away from Selenium.

leeuwe:
@michael-siek I've seen the PR and thanks for the work on it. Would you be able to give a rough estimate on when it will be available? This would be good to have as I need to make a decision how I'm going to implement axe-core, either by doing a bit more work to use axe-core, or wait for axe-core-npm/playwright to be available.
Thank you in advance.

michael-siek:
Hey @leeuwe we are thinking that we will do a public release sometime in June.

padmavemulapati:
Validated @axe-core/playwright Integration, following `ReadME` document.
Everything working as expected.
Steps followed for validation:
1. Installed `Playwright` (playwright@1.12.0)
2. Installed `@axe-core/playwright` (@axe-core/playwright@4.2.2-alpha.135)
3. created a test file 
``` test.js
const AxeBuilder = require('@axe-core/playwright').default;
const playwright = require('playwright');
const fs = require('fs');

(async () => {
  const browser = await playwright.chromium.launch();
  //const builder = new AxeBuilder({ page });
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto('http://qateam.dequecloud.com/attest/api/test.html');

  try {
    const results = await new AxeBuilder({ page }).analyze();
  
    fs.writeFile("./options.json", JSON.stringify(results, null,4),(err)=>
    {if(err){
        console.error(err);
        return;
        };
        console.log("File has been created");
    });
  } catch (e) {
    // do something with the error
  }
  await browser.close();
})();
```
got the expected results. and validated all the remaining APIs, `include` , `exclude`, `disableRules`, `withRules`, `withTags`, `options` etc.,

michael-siek:
Hi @leeuwe, [`@axe-core/playwright`](https://www.npmjs.com/package/@axe-core/playwright) is now available but only as a canary release. We will do a stable release soon!

leeuwe:
[...]

PR #245, opened 2021-05-05
feat(playwright): add playwright integration

michael-siek:
See: #100


packages/playwright/src/test.ts
```diff
@@ -0,0 +1,239 @@
...
+      // Ensure all run rules had the "foobar" tag
+      assert.deepStrictEqual(0, all.length);
+    });
+  });
+
+  describe('iframe tests', () => {
+    it('injects into nested frames', async () => {
+      await page.goto(`${addr}/nested-frames.html`);
+      const results = await new AxeBuilder({ page }).analyze();
+      assert.strictEqual(results.incomplete.length, 0);
```
straker, line 186: This test would be more clear on its purpose and more resilient if the test looked for the `frame-tested` rule in the `passes` results. 

As it stands it was confusing why testing for incomplete would prove it was injected into frames since removing the iframe from the fixture would still pass the test. If there was no iframe on the page the `frame-tested` rule would move to `inapplicable` results

straker (dismissed):

packages/playwright/src/index.ts
```diff
@@ -0,0 +1,166 @@
...
+  /**
+   * Get axe-core source and configurations
+   * @returns String
+   */
+
+  private script(): string {
+    return `
+        ${this.source}
+        axe.configure({ 
+          allowedOrigins: ['<unsafe_all_origins>'],
```
stephenmathieson, line 161: IMO this should be configurable. I know you're just matching the other integration's API, but since this is a _new_ library, I think it makes sense to implement configurable origins here first (and follow up in the other libs).

Also, do we want an "unsafe" option enabled by default in new code?

Probably need input from @WilcoFiers and/or @dylanb here.

straker: For our open source libraries we don't allow customization and should match what axe-core does by default, which is test all iframes. For now this is the appropriate config, but we do have plans for axe-core 4.3 to allow our open source code to not need this anymore and use a different API. When 4.3 lands we will remove these.

stephenmathieson: Thanks for clarifying.

straker (approved):

Issue #230, opened 2021-04-23
Cli stuck, not exiting


marcuslindblom:
I use 4.1.4 and when I test my page the cli just hangs with this message. 

```

Running axe-core 4.1.4 in chrome-headless
Testing https://nimbleinitiatives.com ... please wait, this may take a minute.
```

I tried the verbose flag but nothing happens. Is it possible to debug and see what happens?

marcuslindblom:
It seems to happen when the page does not have any errors

Elte156:
> @Elte156 I also have a CSP in place so I think we have the exact same issue. If I understand correctly this fix just issue the timeout?

Correct @marcuslindblom. The PR will at least exit, which presents the opportunity to give error feedback, instead of just hanging.

I found that if you use the CSP header that I described in the PR, AXE will be compatible with the rest of your CSP and process the page.

marcuslindblom:
Ok, setting the CSP to unsafe is not an option for me so I guess I have to wait for an other fix 👍🏻

Elte156:
@marcuslindblom I understand.

I got around this by having more lenient CSP rule sets in lower environments like local, test, and beta. Then I just test with AXE-cli without issue.

The axe puppeteer package mentions this toggle `setBypassCSP` to overcome the CSP issues.
https://github.com/dequelabs/axe-core-npm/blob/develop/packages/puppeteer/README.md#bypassing-content-security-policy
So I figured I'd sort of do the same during testing.


https://github.com/dequelabs/axe-core-npm/blob/a9e792127a1e373fc7f42e9db4d45d58b6f05930/packages/cli/src/lib/axe-test-urls.ts#L34-L36
I believe this is what a strict CSP header (`script-scr`) does not like and will block AXE from continuing. 


**Are there any more instances where AXE needs to inject itself into the page?**
I wonder if there is an alternative to the script injection, to determine if axe is ready.

PR #250, opened 2021-05-11
fix(cli): add timeout waiting for page to be ready

Elte156:
Partly addresses #230 

I discovered my CLI issue was because of the Content Security Policy (CSP) headers that I set in my web app. 
When I set `Content-Security-Policy` to `script-src 'unsafe-inline';`, AXE is then able to process the page.

This timeout is to force AXE to fail after 10 seconds of waiting; instead of indefinitely waiting for `.deque-axe-is-ready` (which will never be injected because of strict CSP headers in place).

**Sample output with timeout in place**
```console
Running axe-core 4.2.0 in chrome-headless

Testing http://example.com ... please wait, this may take a minute.
 
An error occurred while testing this page.
Please report the problem to: https://github.com/dequelabs/axe-cli/issues/
```

### DIFF
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -41,7 +41,8 @@ const testPages = async (
         return driver.wait(
           WebDriver.until.elementsLocated(
             WebDriver.By.css('.deque-axe-is-ready')
-          )
+          ),
+          10000
         );
       })
       .then(() => {


michael-siek:
Thank you very much for your contribution @Elte156!

michael-siek (approved):

michael-siek:
- [x] reviewed for security

Elte156:
No problem! @michael-siek

PR #253, opened 2021-05-14
chore: sync master with develop

michael-siek:


stephenmathieson (approved):
Should [make a robot do this](https://github.com/dequelabs/walnut/blob/develop/.github/workflows/sync-branches.yml).

PR #254, opened 2021-05-18
fix: update axe-core to 4.2.1

michael-siek:


packages/webdriverjs/package.json
```diff
@@ -76,10 +76,10 @@
     "typescript": "^4.1.5"
   },
   "dependencies": {
-    "axe-core": "^4.2.0"
+    "axe-core": "^4.2.1"
   },
   "peerDependencies": {
-    "selenium-webdriver": ">= 2.53.1"
+    "selenium-webdriver": "<= 4.0.0-beta.3 || >= 2.53.1"
```
stephenmathieson, line 82: Was this change intentional? Seems unrelated.

michael-siek: This was intentional to remove the warning message since we use a `prerelease` version of selenium-webdriver.

stephenmathieson (approved):

packages/webdriverjs/package.json
dbjorge: This might be a little wider than you want; I think the `<= 4.0.0-beta.3` part would accept v1.0.0.

dbjorge: (you can test this locally using the same library npm uses for semver matching with `npx semver -r "<= 4.0.0-beta.3 || >= 2.53.1" "1.0.0"` - it outputs `1.0.0`, which indicates that it believes `1.0.0` matches the given range)

packages/cli/package.json
```diff
@@ -41,7 +41,7 @@
   ],
   "dependencies": {
     "@axe-core/webdriverjs": "^4.2.0",
-    "axe-core": "^4.2.0",
+    "axe-core": "^4.2.1",
```
stephenmathieson, line 44: Just _another_ reminder: users were able to use 4.2.1 before this patch landed. Since we have the `^` here, we'll allowing npm to install a _range_ of versions, not exclusively 4.2.1 (previously 4.2.0).

stephenmathieson (approved):

PR #255, opened 2021-05-18
feat: respect explicit allowedOrigins configurations

dbjorge:
#240 included a change to have the assorted integration libraries configure a default `allowedOrigins` policy, but in doing so, it overrides any explicitly-configured `allowedOrigins` settings configured by a library consumer. We have an `@axe-core/puppeteer` consumer (https://github.com/microsoft/accessibility-insights-service) where we'd like to be able to override the default policy; this PR implements that support by moving the configure call which sets the default to before the configure call with user-provided configuration.

Our team technically only needs this for `@axe-core/puppeteer`, but for consistency's sake I've included similar changes in each of the integration libraries which already included support for arbitrary `configure` calls (puppeteer, react, webdriverjs, but not webdriverio, since it doesn't appear to support an arbitrary configure spec).

We'd really appreciate if this could get in for the 4.2.1 release (@michael-siek , it looks like you're running this?) - we'd like to be able to use `allowedOrigins` as part of picking up v4.2.x if possible.

### DIFF
--- a/packages/puppeteer/test/integration/doc-dylang.test.ts
+++ b/packages/puppeteer/test/integration/doc-dylang.test.ts
@@ -27,7 +27,7 @@ describe('doc-dylang.html', function () {
   before(async function () {
     // const app: express.Application = express()
     const app: express.Application = express();
-    app.use(express.static(path.resolve(__dirname,  '..', 'fixtures')));
+    app.use(express.static(path.resolve(__dirname, '..', 'fixtures')));
     this.server = createServer(app);
     this.addr = await testListen(this.server);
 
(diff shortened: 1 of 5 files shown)


packages/puppeteer/test/integration/doc-dylang.test.ts
```diff
@@ -27,7 +27,7 @@ describe('doc-dylang.html', function () {
   before(async function () {
     // const app: express.Application = express()
     const app: express.Application = express();
-    app.use(express.static(path.resolve(__dirname,  '..', 'fixtures')));
+    app.use(express.static(path.resolve(__dirname, '..', 'fixtures')));
```
dbjorge, line 30: The formatting changes to the existing tests were recommended by `npm run fmt` when I ran it as part of validating the new test

lisli1 (approved):
LGTM

stephenmathieson (commented):
👍 from me. I thought it was odd that we weren't allowing this to be configurable [elsewhere](https://github.com/dequelabs/axe-core-npm/pull/245#discussion_r628701436). Will need an OK from Micheal tho.

WilcoFiers:
We are working on [features for axe-core 4.3](https://github.com/dequelabs/axe-core/issues/2936) which will make allowedOrigins irrelevant for Selenium / Puppeteer.  We did not add `allowedOrigins` as a configurable option because we would have to remove it again once axe-core 4.3 came out. That would be a breaking change that would require us to go to 5.0, which we can not do for a variety of reasons.

For axe-core 4.2 there are a few options available to let you configure what frames run. You can exclude specific frames, or all frames using the `exclude` option. You can also use the `iframes` option, set it to `false`, which turns off all frame testing. And to prevent less information getting passed between frames there is the `noHtml` option.

dbjorge:
Thanks for sharing that, Wilco! We'll pursue a workaround until the 4.3 changes are available.

PR #258, opened 2021-05-19
fix: selenium-webdriverjs peer dependency to allow newer versions

michael-siek:


packages/webdriverjs/package.json
```diff
@@ -79,7 +79,7 @@
     "axe-core": "^4.2.1"
   },
   "peerDependencies": {
-    "selenium-webdriver": ">= 2.53.1 <=4.0.0-beta.3"
+    "selenium-webdriver": ">3.0.0-beta  || >=2.53.1 || >4.0.0-alpha"
```
stephenmathieson, line 82: What versions are you targeting here? This seems more complicated than it needs to be.

michael-siek: This should allow users to use any version from 2.53.1+, 3.0.0-beta+, and 4.0.0-alpha+

WilcoFiers (approved):

PR #260, opened 2021-05-25
Add configuration option to optional disable cache (deduplication)

### LINKED ISSUE #256: React - feature proposal: Optionally turn off deduplication
Product: react

Expectation: 
Would love the ability to disable deduplication via a config option.

Actual: 
Currently the @axe-core/react stores reported issues in a cache (via key). If that issue has appeared before it will filter it out the logged issues.

Motivation: 
To give a little context on why I would like this feature, we currently invoke the axe tool when either of the following occur:
1. on initial page load
2. whenever a hot module replacement occurs
3. whenever the developer changes pages (via React router)

For case 1 everything is working perfectly. However, in case 2 and 3 I would love for all issues to be printed to the console. The reason for this is as follows. Imagine someone going through and fixing issues. They make changes to the code, go back to the browser and see if they fixed the issue. Hot module replacement kicks in and only replaces the changed module. (without refreshing the page). 

Since only "new" issues are printed to the console, the developer can't really determine if they resolved the issue. They now need to refresh the entire page to see if the issue has been resolved. For step 3, the same idea happens. Whenever the developer navigates to a different page (with React router) only "new" issues are printed to the console even if we're on an entirely different page. The benefit here is that you can treat each page change as an entire refresh, allowing you to easily determine which pages have issues and which do not.

From what I gather in the code, this is a simple addition that I can get in a PR tomorrow. If you take a look at [this function](https://github.com/dequelabs/axe-core-npm/blob/cbb795f1a92c419794a5f1f9645e28493d7c9bdb/packages/react/index.ts#L216), you will notice that it first checks to see if the issue's key exists in a cache object. If it does exist, the issue is filtered out. If it does not exist, it passes through the filter.

My proposal here is that we add a new configuration option next to `runOnly` called `deduplicate`. By default this will remain `true` as an opt-in feature/change. If this property is `false` then the filter function will always return `true` thus outputting all issues to console.

Let me know if this makes sense :).

<pre><code>
axe-core version: 4.1.1
@axe-core/react: 4.1.1

- Node version: 12.14.0
- Platform: all?
</code></pre>

charle692:
Closes #256

As mentioned in the issue, would love a way to disable the deduplication. I've detailed why and how this would be useful in the linked issue.

There's a few different ways that this can be done. I've decided on this solution but I'm open to suggestions :).
I thought of clearing the cache outright if `disableCache` is present but I thought it would be better to just ignore the cache instead as it would be less disruptive to other potential workflows? This way you can remove the option and take advantage of anything cached on previous runs.

Would really like to get this merged in, if there's anything missing or something I've done wrong, please let me know!


charle692:
Just messaging to see if anyone has had a chance to look at this? We are starting an accessibility project soon and this would really help solve some of the workflow issues we are experiencing.

michael-siek:
Hi @charle692, do you mind providing a test for this option.

charle692:
Hi @michael-siek, sorry about not including specs in the first place. I took some time to attempt at writing some specs but I believe I'm running into an issue with the test environment.

What I mean by that is the **axe** call in **example.js** is interfering with the specs and causing false positives. If we comment out that code as such:

image

Rebuild the example, you will see that whenever you run the specs, all previous specs fail. The reason for this is that the **axe** call inside of the specs doesn't report anything to the spies. The only calls that are recorded to the **groupCollapsed** are those from the call in **example.js**.

With the call to axe in the example removed, I added a debugger to the following spec:
image

Which allows me to verify if **groupCollapsed** has been called. As you can see in the below screenshot, it reports as never being called even if we are invoking **axe** from within the spec.

image

So what this means is that the **axe** calls in the specs are not influencing the spies and so we aren't actually testing anything properly. 

I tried to write a test like so, that would verify on the second call to **axe**, **groupCollapsed** would be invoked again indicating that issues are printed to console rather than being deduplicated.
```js
it('should duplicate logs when cache is disabled', function (done) {
    cy.visit('http://localhost:8080').then(function (win) {
      const groupCollapsed = cy.spy(win.console, 'groupCollapsed');
      cy.document().shadowGet('#service-chooser').shadowFirst();

      axe(React, ReactDOM, 0, { disableCache: true }).then(function () {
        expect(groupCollapsed.args).not.to.eql([]);
        groupCollapsed.reset();

        axe(React, ReactDOM, 0, { disableCache: true }).then(function () {
          expect(groupCollapsed.args).not.to.eql([]);
          done();
        });
      });
    });
  });
```

I've also attempted to write a test that would verify that the caching is still working whenever it's not disabled. And same issue here due to **axe** in the specs not influencing the spies.
```js
it('should not duplicate logs when cache is enabled', function (done) {
    cy.visit('http://localhost:8080').then(function (win) {
      const groupCollapsed = cy.spy(win.console, 'groupCollapsed');
      cy.document().shadowGet('#service-chooser').shadowFirst();

      axe(React, ReactDOM, 0, {}).then(function () {
        expect(groupCollapsed.args).not.to.eql([]);
        groupCollapsed.reset();

        axe(React, ReactDOM, 0, {}).then(function () {
          expect(groupCollapsed.args).to.eql([]);
          done();
        });
      });
    });
  });
```

So all that to say that a test for this can't be accurately written until we can figure out a way to have calls to **axe** report to spies properly. As it is, all the specs that are currently written are not getting tested properly. Hopefully that is clear enough, if not then I can try and clarify further.

packages/react/index.ts
```diff
@@ -217,7 +220,7 @@ function checkAndReport(node: Node, timeout: number): Promise<void> {
               const key: string = node.target.toString() + result.id;
               const retVal = !cache[key];
               cache[key] = key;
-              return retVal;
+              return disableCache || retVal;
```
WilcoFiers, line 223: This isn't really preventing the cache from being used, as much as it's just overriding it. Probably better to just prevent it from even calling `violations.filter`.

packages/react/index.ts
```diff
@@ -334,6 +337,7 @@ function addComponent(component: any): void {
  */
 interface ReactSpec extends axeCore.Spec {
   runOnly?: string[];
+  disableCache?: boolean;
```
WilcoFiers, line 340: Please describe this option in the readme.

packages/react/index.ts
```diff
@@ -186,6 +187,8 @@ function failureSummary(
  * @return {Promise}
  */
 function checkAndReport(node: Node, timeout: number): Promise<void> {
+  const disableCache = conf['disableCache'];
```
WilcoFiers, line 190: I'm all for this, but would suggest this be called `disableDeduplicate`.

```suggestion
  const disableCache = conf['disableDeduplicate'];
```

WilcoFiers (changes requested):

packages/react/index.ts
charle692: Yeah so I sort of did this intentionally. I was thinking that some invocations of the library might want to leverage cache from previous runs?

Example, a hot reload would ignore cache and do a full accessibility test to verify that an issue was fixed properly but secondary data load from a page might only report on new errors. Does that make sense?

packages/react/index.ts
charle692: Yup, good catch will do :)

packages/react/index.ts
charle692: Yup I can make that change :)

packages/react/index.ts
WilcoFiers: Ahhh, I see what you mean. I like it.

charle692:
@WilcoFiers sorry about the delay, had to temporarily switch priorities for a bit. How do the changes look?

michael-siek:
Hey @charle692 do you mind resolving merge conflicts when possible. It looks like your branch may be out of date from `develop`.

charle692:
@michael-siek I've fixed the conflicts. I noticed a few changes in my MR due to the formatting tools kicking in on commit due to husky. I tried fixing them but realized afterwards that this is something configured by the project itself. 

I'm not sure if you want me to disable that somehow to clean up the diff? Seems like some previous changes didn't pass through the formatted somehow.

michael-siek:
Hi @charle692 everything LGTM just going to get this running in circleci and we will have it merged.

straker:
@charle692 So we've had to cherry-pick your commits into a new branch under my name to get CircleCi to run properly. We're going to merge that in and then close this one.

charle692:
> @charle692 So we've had to cherry-pick your commits [into a new branch](https://github.com/dequelabs/axe-core-npm/pull/309) under my name to get CircleCi to run properly. We're going to merge that in and then close this one.

No worries, thanks for the heads up and thanks for merging it!

PR #263, opened 2021-06-05
feat: update `axe-core@4.2.2`

michael-siek:


packages/playwright/package.json
```diff
@@ -10,6 +10,9 @@
   "license": "MPL-2.0",
   "main": "dist/index.js",
   "typings": "dist/index.d.ts",
+  "publishConfig": {
+    "access": "public"
```
WilcoFiers, line 14: I'm not sure what this does, but looking at the docs "public" is for unscoped packages, which this is not.

WilcoFiers (changes requested):

packages/playwright/package.json
stephenmathieson: This is needed, unless we're going to explicitly set it during `npm publish`. This is probably why `@axe-core/playwright` is not publicly available still.

stephenmathieson: From [npm docs](https://docs.npmjs.com/cli/v7/using-npm/config#access):

> Default: 'restricted' for scoped packages, 'public' for unscoped packages

WilcoFiers (approved):

PR #264, opened 2021-06-07
fix(playwright): Make package public

michael-siek:
See: https://github.com/dequelabs/axe-core-npm/pull/263#discussion_r646680440


stephenmathieson (approved):
Change LGTM but I don't think the commit message correctly describes it.

Maybe:

```
fix(playwright): Make package public
```

PR #274, opened 2021-06-17
fix(puppeteer): test frames in shadow trees [WIP]

WilcoFiers:
I suspect this problem exists in other APIs in this repo too. Don't yet know how to make the disableFrame thing work, we'll have to work on this further.


packages/puppeteer/test/integration/shadow-dom.test.ts
```diff
@@ -0,0 +1,53 @@
+// Adapter from axe-webdriverjs.
```
michael-siek, line 1:
```suggestion
// Adapter from axe-puppeteer.
```

michael-siek (changes requested):

PR #297, opened 2021-07-06
[wip]Fix typings for current versions of WebdriverIO

cunderw:
The current typings do not work with WebdriverIO and should be updated to current supported typings.

### DIFF
--- a/packages/webdriverio/src/types.ts
+++ b/packages/webdriverio/src/types.ts
@@ -1,23 +1,19 @@
 import type { AxeResults, ElementContext, RunOptions, Spec } from 'axe-core';
 import * as axe from 'axe-core';
 import {
-  BrowserObject as BrowserObjectSync,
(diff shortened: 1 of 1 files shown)


straker:
This is a breaking change isn't it? By removing typings from v6 of webdriver and moving to v7 typings, we break the ability to support v6 webdriverio using typescript don't we?

cunderw:
> This is a breaking change isn't it? By removing typings from v6 of webdriver and moving to v7 typings, we break the ability to support v6 webdriverio using typescript don't we?

It would indeed break support for v6 which is in LTS support until Feb. 2022 before becoming entirely deprecated.

michael-siek:
Hi @cunderw we are currently discussing the best way to support new typings without breaking changes as well as supporting webdriverio v7 (without dropping support for v5-6) closing this PR until we have come up with a solution.

PR #298, opened 2021-07-07
feat: use axe.runPartial instead of axe.run

WilcoFiers:


AdnoC:
Seems reasonable to me. Will have to look into working with shadow elements with selenium, as a cursory search reveals that they might not be given first class support.

PR #312, opened 2021-07-30
feat(webdriverjs): upgrade to axe-core 4.3

WilcoFiers:


packages/webdriverjs/src/browser.ts
```diff
@@ -0,0 +1,193 @@
...
+        });
+        var config = ${JSON.stringify(config)};
+        if (config) {
+          window.axe.configure(config);
+        }
+        return (
+          window &&
+          window.axe &&
+          typeof window.axe.runPartial === 'function'
+        )
```
WilcoFiers, line 39:
```suggestion
        return (typeof window.axe.runPartial === 'function')
```

packages/webdriverjs/src/test/test.ts
```diff
@@ -154,35 +187,76 @@ describe('@axe-core/webdriverjs', () => {
     });
   });
 
-  describe('iframe tests', () => {
+  describe('frame tests', () => {
```
WilcoFiers, line 190:
```suggestion
  describe('frameset tests', () => {
```

packages/webdriverjs/src/browser.ts
```diff
@@ -0,0 +1,193 @@
...
+    driver.executeScript<boolean>(`
+      return (
+        window &&
+        window.axe &&
+        typeof window.axe.runPartial === 'function'
+      )
+    `)
+  );
+}
+
```
straker, line 136: This is no longer needed.

```suggestion
```

packages/webdriverjs/src/index.ts
```diff
@@ -1,9 +1,20 @@
...
+import { normalizeContext, sleep } from './utils/index';
 import AxeInjector from './axe-injector';
+import { AxePartialRunner, PartialResults } from './axe-partial-runner';
+import {
+  axeGetFrameContext,
+  axeRunPartial,
+  axeRunLegacy,
+  axeSourceInject,
+  axeFinishRun,
+  axeSupportsRunPartial,
```
straker, line 14: I'm surprised ESLint didn't catch the unused import.

```suggestion
```

packages/webdriverjs/src/browser.ts
```diff
@@ -0,0 +1,193 @@
...
+/**
+ * Selenium-webdriver thenable aren't chainable. This fixes it.
+ */
+function promisify<T>(thenable: Promise<T>): Promise<T> {
+  return new Promise((resolve, reject) => {
+    thenable.then(out => {
+      if (isSerialError(out)) {
+        // Throw if we find an error-like object
+        reject(new Error(out.stack));
```
straker, line 174: I know that once the promise is `rejected` the `resolve` will noop but we should still stop the execution here to prevent any potential bugs later if we add more to the outside of this if statement

```suggestion
        return reject(new Error(out.stack));
```

straker (dismissed):

straker (approved):

packages/webdriverjs/src/test/test.ts
```diff
@@ -154,35 +187,76 @@ describe('@axe-core/webdriverjs', () => {
...
+    it('reports erroring frames in frame-tested', async () => {
+      await driver.get(`${addr}/crash-me-parent.html`);
+      const results = await new AxeBuilder(driver, axeSource + axeCrasherSource)
+        .options({ runOnly: ['label', 'frame-tested'] })
+        .analyze();
+
+      assert.equal(results.incomplete[0].id, 'frame-tested');
+      assert.lengthOf(results.incomplete[0].nodes, 1);
+      assert.equal(results.violations[0].id, 'label');
+      assert.lengthOf(results.violations[0].nodes, 2);
```
straker, line 259: We should check the selector of these is correct

straker (changes requested):

straker (dismissed):

straker:
So I am not a fan of [the double function recursion](https://github.com/dequelabs/axe-core-npm/pull/312/files#diff-ff14440feeef60d9e6faeb51c65ada3c16684ef06382281c5ca9877de1b49d4fR177-R227). I find it confusing and hard to follow exactly how it works (such as error handling and what the final result is).

However, I'd like to ask @nloding and @michael-siek for their opinion on the code since you two are the maintainers. If you find the approach acceptable I can approve the PR.

michael-siek:
I agree with @straker I think it would be easier to read with a single recursive function since a developer would have to go back and forth to get a full understanding of what is going on.

nloding:
I'm in agreement. Since we've learned that it's all synchronous regardless, it makes the code harder to follow and, IMO, is tech debt.

straker (dismissed):
Please refactor the double function recursion.

straker (approved):

PR #327, opened 2021-08-12
feat(puppeteer): Upgrade to axe-core 4.3

WilcoFiers:


packages/puppeteer/src/axePartialRunner.ts
```diff
@@ -0,0 +1,53 @@
+import * as axe from 'axe-core';
+
+export type PartialResults = Parameters<typeof axe.finishRun>[0];
+
+/**
+ * This class parallelizes the async calls to axe.runPartial.
+ *
+ * In this project, most async calls needs to block execution, such as
+ * the driver.switchTo().frame() calls. These must run in the correct order
+ * to avoid stale element references and infinite loops testing the same frame.
```
WilcoFiers, line 10: Oops, copy from WDJS. Will fix.

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -0,0 +1,711 @@
...
+          run: () => new Promise(resolve => resolve({})),
+          configure: () => {}
+        }
+      `;
+      await page.goto(`${addr}/external/index.html`);
+      const evalSpy: SinonSpy = sinon.spy(page.mainFrame(), 'evaluate');
+      await new AxePuppeteer(page, axeSource).analyze();
+      assert(evalSpy.calledWith(axeSource));
+    });
+    // TODO: Defaults to using the bundled axe-core source
```
straker, line 108: Do we need to worry about this TODO in this pr?

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -0,0 +1,711 @@
...
+            tags: ['wcag2aa']
+          }
+        ]
+      };
+
+      await page.goto(`${addr}/external/index.html`);
+
+      // HACK: work around axe-core (incorrectly) requiring this to be
+      // a function (see https://github.com/dequelabs/axe-core/issues/974).
+      (config.checks as any)[0].evaluate = 'function () { return false }';
```
straker, line 472: I believe this should no longer be required as evaluate can be a string [in the typings](https://github.com/dequelabs/axe-core/blob/develop/axe.d.ts#L213)

```suggestion
      config.checks[0].evaluate = 'function () { return false }';
```

straker (approved):
Love all the deleted files. I'm fine leaving the hack in-place for this pr as it's just a typescript hack that isn't needed anymore.

packages/puppeteer/test/axePuppeteer.test.ts
WilcoFiers: Was copied from the other source... probably shouldn't be here. I'll just take it out.

packages/puppeteer/test/axePuppeteer.test.ts
WilcoFiers: Copied this from the previous implementation, but yeah it can be removed.

PR #331, opened 2021-08-13
feat(webdriverio): Upgrade to axe-core@4.3.3

michael-siek:
Blocked by: https://github.com/dequelabs/axe-core-npm/pull/333


packages/webdriverio/src/utils/index.ts
```diff
@@ -104,3 +82,116 @@ export const logOrRethrowError = (error: Error): void => {
...
+        branding: { application: 'webdriverio' }
+      });
+
+      var runPartial = typeof window.axe.runPartial === 'function';
+      return { runPartialSupported: runPartial };
+    `)
+  );
+};
+
+export const axeRunPartialSupported = (
```
WilcoFiers, line 113: This is never used.

WilcoFiers (changes requested):

packages/webdriverio/src/index.ts
```diff
@@ -264,4 +288,59 @@ export default class AxeBuilder {
...
     }
   }
+
+  /**
+   * Get partial results from the current context and its child frames
+   * @param {ContextObject} context
+   * @param {Boolean} initiator - defaults to false
+   */
+
+  private async runPartialRecursive(
```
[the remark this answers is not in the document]
michael-siek: extending webdriverjs broke the original webdriverio package which is why we went this route. I can write a proposal so we could just extend a class for a lot less copy and paste if needed.

packages/webdriverio/src/index.ts
```diff
@@ -222,6 +218,34 @@ export default class AxeBuilder {
...
+    }
+    const partials = await this.runPartialRecursive(context, true);
+
+    return await this.finishRun(partials);
+  }
+
+  private async runLegacy(context: ContextObject): Promise<AxeResults> {
+    const { client, option } = this;
+    await this.inject();
+    return axeRunLegacy({ client, context, options: option });
```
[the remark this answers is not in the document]
michael-siek: `this.options` is chainable and therefore you cannot have `this.options` twice

packages/webdriverio/src/test.ts
```diff
@@ -108,311 +118,559 @@ describe('@axe-core/webdriverio', () => {
...
+            // Ensure all run rules had the "foobar" tag
+            assert.deepStrictEqual(0, all.length);
+          });
         });
 
-        it('with only include', async () => {
-          let error: Error | null = null;
-          await client.url(`${addr}/context.html`);
-          const builder = new AxeBuilder({ client }).include('.include');
+        describe('include/exclude', () => {
```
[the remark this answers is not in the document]
michael-siek: We can open a single tech debt ticket with task to remove these (if one hasn't already been opened)

packages/webdriverio/src/index.ts
```diff
@@ -1,18 +1,23 @@
 import type { RunOptions, AxeResults } from 'axe-core';
-import { source } from 'axe-core';
+import { source, ContextObject } from 'axe-core';
```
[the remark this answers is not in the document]
stephenmathieson: > JavaScript no longer guarantees Function.prototype.toString() returning the complete source.

Can you provide your source for this information?

packages/webdriverio/src/test.ts
```diff
@@ -108,311 +118,559 @@ describe('@axe-core/webdriverio', () => {
...
+              .analyze();
+
+            const all = [
+              ...results.passes,
+              ...results.inapplicable,
+              ...results.violations,
+              ...results.incomplete
+            ];
+            // Ensure all run rules had the "foobar" tag
+            assert.deepStrictEqual(0, all.length);
```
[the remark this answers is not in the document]
michael-siek: Didn't know this was a thing this is pretty cool.

packages/webdriverio/src/test.ts
```diff
@@ -108,311 +118,559 @@ describe('@axe-core/webdriverio', () => {
...
-      describe('analyze', () => {
-        it('returns results', async () => {
-          await client.url(`${addr}/index.html`);
-          const results = await new AxeBuilder({ client }).analyze();
-          assert.isNotNull(results);
-          assert.isArray(results.violations);
-          assert.isArray(results.incomplete);
-          assert.isArray(results.passes);
-          assert.isArray(results.inapplicable);
+      describe('axe-core@4.0.3', () => {
```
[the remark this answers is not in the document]
michael-siek: This integration isn't able to be configured

packages/webdriverio/package.json
```diff
@@ -33,7 +33,8 @@
     "webdriverio"
   ],
   "dependencies": {
-    "axe-core": "^4.2.3",
+    "axe-core": "^4.3.2",
+    "axe-test-fixtures": "github:dequelabs/axe-test-fixtures",
```
stephenmathieson, line 37: Should be a dev dep

stephenmathieson (dismissed):

packages/webdriverio/src/index.ts
```diff
@@ -182,7 +179,6 @@ export default class AxeBuilder {
     return `
       ${this.axeSource}
       axe.configure({ 
-        allowedOrigins: ['<unsafe_all_origins>'],
```
WilcoFiers, line 185: We should keep these around for axe-core 4.2 users IMO.

packages/webdriverio/src/test.ts
```diff
@@ -3,7 +3,6 @@ import * as webdriverio from 'webdriverio';
 const sync = require('@wdio/sync').default;
 import * as wdio from '@wdio/sync';
 import * as express from 'express';
-import * as sinon from 'sinon';
```
WilcoFiers, line 6: Take it out of package.json too please.

packages/webdriverio/src/test.ts
```diff
@@ -63,14 +62,24 @@ describe('@axe-core/webdriverio', () => {
...
     let client: webdriverio.BrowserObject;
-    const axeSourcePath = path.resolve(
-      './fixtures/external/axe-core@legacy.js'
+    const axePath = require.resolve('axe-core');
+    const axeSource = fs.readFileSync(axePath, 'utf8');
+    const axeTestFixtures = path.resolve(
+      __dirname,
+      '..',
+      'fixtures',
+      'external'
```
WilcoFiers, line 71: I'm not sure how this is passing CI, but I don't see the fixture files in package.json. Am I missing something?

packages/webdriverio/src/test.ts
```diff
@@ -425,16 +714,28 @@ describe('@axe-core/webdriverio', () => {
...
+
+        it('returns an error as the first argument', async () => {
+          await client.url(`${addr}/index.html`);
+          new AxeBuilder({ client, axeSource: 'throw new Error()' }).analyze(
+            (err, results) => {
+              try {
+                assert.isNull(results);
+                assert.isNotNull(err);
+              } catch (e) {
+                // Do something here ¯\_(ツ)_/¯
```
WilcoFiers, line 734: You'll have to use promise change to make this work. You'll have to change this to be:

```suggestion
                done(e)
```

Otherwise this test can never fail.

packages/webdriverio/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  return true;
+};
+
+/**
+ * Get running context
+ * @param {Array} include
+ * @param {Array} exclude
+ * @returns {(ElementContext | null)}
```
WilcoFiers, line 43: See this is why I don't think we should document our params / returns. If we don't pay close attention to them they go out of sync with TypeScript.

```suggestion
 * @returns {ElementContext}
```

WilcoFiers (changes requested):

packages/webdriverio/src/test.ts
stephenmathieson: They're a root-level dependency, see https://github.com/dequelabs/axe-core-npm/blob/develop/package.json#L49

packages/webdriverio/src/test.ts
```diff
@@ -723,15 +723,16 @@ describe('@axe-core/webdriverio', () => {
           });
         });
 
-        it('returns an error as the first argument', async () => {
-          await client.url(`${addr}/index.html`);
+        it('returns an error as the first argument', done => {
+          client.url(`${addr}/index.html`);
```
WilcoFiers, line 727: You need to wait for this to load. Otherwise this test doesn't prove what you're trying to.

```suggestion
          client.url(`${addr}/index.html`);
          .then(() => 
```

WilcoFiers (changes requested):

packages/webdriverio/src/test.ts
michael-siek: test fixed had to add a `// @ts-ignore` due to this error message
```console
TSError: ⨯ Unable to compile TypeScript:
src/test.ts:727:44 - error TS2339: Property 'then' does not exist on type 'void'.

727           client.url(`${addr}/index.html`).then(() => {
```

packages/webdriverio/src/test.ts
```diff
@@ -724,18 +724,21 @@ describe('@axe-core/webdriverio', () => {
...
-          new AxeBuilder({ client, axeSource: 'throw new Error()' }).analyze(
-            (err, results) => {
-              try {
-                assert.isNull(results);
-                assert.isNotNull(err);
-                done();
-              } catch (e) {
-                done(e);
+          // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+          // @ts-ignore
```
WilcoFiers, line 728: Wouldn't we be better off putting the wdio/sync stuff into its own test file, so we're not fighting the type system here?

packages/webdriverio/src/index.ts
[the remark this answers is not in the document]
stephenmathieson: From https://tc39.es/Function-prototype-toString-revision/#proposal-sec-function.prototype.tostring:

> If func is a Bound Function exotic object or a built-in Function object, then return an implementation-dependent String source code representation of func. The representation must have the syntax of a NativeFunction. Additionally, if func is a Well-known Intrinsic Object and is not identified as an anonymous function, the portion of the returned String that would be matched by PropertyName must be the initial value of the name property of func.

If I'm reading this correctly, `Function#toString()` is fine for our use case.

packages/webdriverio/src/test.ts
```diff
@@ -111,13 +120,170 @@ describe('@axe-core/webdriverio', () => {
...
+            ]);
+            assert.deepEqual(nodes[2].target, [
+              '#ifr-bar',
+              '#bar-baz',
+              'input'
+            ]);
+            assert.deepEqual(nodes[3].target, ['#ifr-baz', 'input']);
+          });
+
+          it('does not inject into disabled frames', async () => {
```
straker, line 217:
```suggestion
          it('does not return results from disabled framesets', async () => {
```

packages/webdriverio/src/test.ts
```diff
@@ -111,13 +120,170 @@ describe('@axe-core/webdriverio', () => {
...
+              'input'
+            ]);
+            assert.deepEqual(nodes[2].target, [
+              '#frm-bar',
+              '#bar-baz',
+              'input'
+            ]);
+          });
+
+          it('does not error when disabled frame does not exist', async () => {
```
straker, line 247:
```suggestion
          it('does not error when disabled frameset does not exist', async () => {
```

packages/webdriverio/src/test.ts
```diff
@@ -138,45 +304,145 @@ describe('@axe-core/webdriverio', () => {
...
+            'input'
+          ]);
+          assert.deepEqual(nodes[1].target, ['#ifr-foo', '#foo-baz', 'input']);
+          assert.deepEqual(nodes[2].target, ['#ifr-bar', '#bar-baz', 'input']);
+          assert.deepEqual(nodes[3].target, ['#ifr-baz', 'input']);
         });
-      });
 
-      describe('disableFrame', () => {
         it('does not inject into disabled frames', async () => {
```
straker, line 409:
```suggestion
        it('does not inject into disabled frameset', async () => {
```

packages/webdriverio/src/test.ts
```diff
@@ -138,45 +304,145 @@ describe('@axe-core/webdriverio', () => {
...
+            '#frm-foo',
+            '#foo-bar',
+            '#bar-baz',
+            'input'
+          ]);
+          assert.deepEqual(nodes[1].target, ['#frm-foo', '#foo-baz', 'input']);
+          assert.deepEqual(nodes[2].target, ['#frm-bar', '#bar-baz', 'input']);
         });
 
         it('does not error when disabled frame does not exist', async () => {
```
straker, line 428:
```suggestion
        it('does not error when disabled frameset does not exist', async () => {
```

packages/webdriverio/src/utils.ts
```diff
@@ -0,0 +1,189 @@
...
+
+export const axeFinishRun = ({
+  client,
+  axeSource,
+  partialResults,
+  options
+}: AxeFinishRunParams): Promise<AxeResults> => {
+  return promisify(
+    client.executeAsync(`
+    var callback = arguments[arguments.length - 1];
```
straker, line 174: Needs to be updated to use the new `window.open('_blank')` technique

straker (changes requested):

packages/webdriverio/src/utils.ts
```diff
@@ -163,12 +163,29 @@ export const axeRunLegacy = ({
...
+  );
+};
+
+// export const switchToAboutBlank = (client: BrowserObject) => {
+//   return promisify(
+//     client.switchToWindow(
+//       client.getWindowHandle()
+//     )
+//   )
+// }
```
straker, line 180:
```suggestion
```

packages/webdriverio/src/utils.ts
```diff
@@ -163,12 +163,29 @@ export const axeRunLegacy = ({
...
+
 export const axeFinishRun = ({
   client,
   axeSource,
   partialResults,
   options
 }: AxeFinishRunParams): Promise<AxeResults> => {
+  // const originalWindow = client.getWindowHandle()
```
straker, line 188:
```suggestion
```

packages/webdriverio/src/utils.ts
```diff
@@ -163,12 +163,29 @@ export const axeRunLegacy = ({
   );
 };
 
+export const openAboutBlank = (client: BrowserObject) => {
+  return promisify(
+    client.execute(`
+      window.open('about:blank', '_blank');
```
straker, line 169: This will fail the new test fixture test. Also need to add the test fixture test.

straker (changes requested):

WilcoFiers (approved):

PR #335, opened 2021-08-17
feat(playwright): allow `AxeBuilder` to use different version of axe-core

michael-siek:


packages/playwright/src/index.ts
```diff
@@ -10,12 +10,14 @@ export default class AxeBuilder {
   private excludes: string[];
   private option: RunOptions;
   private source: string;
-  constructor({ page }: AxePlaywrightParams) {
+  constructor({ page, axeSource }: AxePlaywrightParams) {
+    const axePath = require.resolve('axe-core');
+    const source = fs.readFileSync(axePath, 'utf-8');
```
WilcoFiers, line 15: Hmm, probably don't need to go to disk if `axeSource` is defined. I know we did that in other projects. I'll leave it up to you.

WilcoFiers (approved):

PR #339, opened 2021-08-25
feat(puppeteer): Deprecate Frame constructors & Puppeteer < 3.0.3

WilcoFiers:
- Upgrade to axe-core 4.3.3
- Call finishRun in a blank page, ensuring environment data exists
- Deprecate construction with Frame, since Puppeteer has no way to create a new page just from a Frame object
- Deprecate Puppeteer < 3.0.2, since Page.browser() is not available in those versions

When either deprecated situation occurs, axe-core/puppeteer will be unable to test cross-origin iframes. These are reported as issues in the frame-tested rule.


packages/puppeteer/src/legacy.ts
```diff
@@ -51,9 +52,8 @@ export async function injectJS(
 }
 
 async function injectJSModule(frame: Frame): Promise<void> {
-  await frame.addScriptTag({
-    path: require.resolve('axe-core')
-  });
+  const source = fs.readFileSync(require.resolve('axe-core'), 'utf8');
+  await injectJSSource(frame, source);
```
WilcoFiers, line 56: I had to make this change, because apparently `addScriptTag` changes the source of the page, which caused problems with the `html` property on results.

straker (approved):

PR #356, opened 2021-09-08
feat(wdjs): Add .setLegacyMode

WilcoFiers:


packages/webdriverjs/package.json
```diff
@@ -63,7 +63,7 @@
     "@types/selenium-webdriver": "^4.0.15",
     "@types/sinon": "^9.0.10",
     "@types/test-listen": "^1.1.0",
-    "axe-test-fixtures": "github:dequelabs/axe-test-fixtures#v1",
+    "axe-test-fixtures": "github:dequelabs/axe-test-fixtures#cross-origin",
```
WilcoFiers, line 66: Awaiting https://github.com/dequelabs/axe-test-fixtures/pull/11

packages/playwright/src/index.ts
```diff
@@ -122,6 +124,18 @@ export default class AxeBuilder {
...
   }
 
+  /**
+   * Use frameMessenger with <same_origin_only>
+   *
+   * This disables use of axe.runPartial() which is called in each frame, and
+   * axe.finishRun() which is called in a blank page. This uses axe.run() instead,
+   * but with the restriction that cross-origin frames will not be tested.
+   */
+  public setLegacyMode(legacyMode = true): this {
```
michael-siek, line 134:
```suggestion
  public setLegacyMode(legacyMode = true): AxeBuilder {
```

packages/playwright/src/index.ts
```diff
@@ -169,12 +183,12 @@ export default class AxeBuilder {
...
     return `
-        ${this.source}
-        axe.configure({ 
-          allowedOrigins: ['<unsafe_all_origins>'], 
-          branding: { application: 'playwright' }
-        })
-        `;
+      ${this.source}
+      axe.configure({
+        ${this.legacyMode ? '' : 'allowedOrigins: ["<unsafe_all_origins>"],'}
```
michael-siek, line 188:
```suggestion
        ${this.legacyMode ? ''allowedOrigins: ["<same_origin>"]' : 'allowedOrigins: ["<unsafe_all_origins>"],'}
```

michael-siek (changes requested):

packages/playwright/src/index.ts
WilcoFiers: That's exactly the same thing but with more characters.

packages/playwright/src/index.ts
WilcoFiers: Should be `this`. If we extend this class (which we do in devtools), `AxeBuilder` is actually wrong.

michael-siek (approved):

PR #358, opened 2021-09-08
chore: add error handling to `finishRun()`

Zidious:
Add error handling to `finishRun()`:

- [ ] playwright
- [ ] puppeteer
- [ ] webdriverio
- [ ] webdriverjs


packages/playwright/src/index.ts
```diff
@@ -147,7 +147,18 @@ export default class AxeBuilder {
...
+
+    try {
+      return this.finishRun(partials);
+    } catch (error) {
+      throw new Error(
+        `Error: ${JSON.stringify(
+          error,
+          null,
+          2
+        )}\n Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md`
```
stephenmathieson, line 159: I don't think this does what you expect it to.

```
throw new Error(`Error: ${JSON.stringify(new Error("boom"), null, 2)}\n Please ...`)
VM335:1 Uncaught Error: Error: {}
 Please ...
    at <anonymous>:1:7
```

Note `Error: {}`, not `Error: boom`.

packages/playwright/src/index.ts
```diff
@@ -259,6 +270,9 @@ export default class AxeBuilder {
         partialResults,
         options
       })
+      .catch(e => {
+        return e;
```
stephenmathieson, line 274: Why `return` the error? Doesn't this mean it's "dropped"?

stephenmathieson: This means the Promise resolves with an instance of an `Error`, thus breaking _all_ type contracts.

```
var result = await new Promise((resolve, reject) => reject(new Error('boom'))).catch(error => { return error })
result instanceof Error
// => true
result
// => Error: boom
    at result (<anonymous>:1:60)
    at new Promise (<anonymous>)
    at <anonymous>:1:20
```

stephenmathieson (changes requested):
`JSON.stringify()`ing an error doesn't do what you're expect it to. Also `return`ing errors breaks type contracts.

packages/playwright/src/index.ts
Zidious: Thoughts on simply throwing the `error` and providing the link to the docs. 


```
   try {
      return this.finishRun(partials);
    } catch (error) {
      throw new Error(
        `Error: ${error}\n Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md`
      );
    }
 ```

File: packages/playwright/src/index.ts [hunk 3]
```diff
@@ -147,7 +147,14 @@ export default class AxeBuilder {
       context
     );
     const partials = await partialResults.getPartials();
-    return this.finishRun(partials);
+
+    try {
+      return await this.finishRun(partials);
+    } catch (error) {
+      throw new Error(
+        `Error: ${error}\n Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md`
```
WilcoFiers, line 155: This won't serialise properly. If you want just the message use error.message, or if you want the stack trace use error.trace.

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -694,6 +694,21 @@ describe('AxePuppeteer', function () {
     });
   });
 
+  describe('axe.finishRun errors', () => {
```
WilcoFiers, line 697: Can you adopt the tests as written?

https://github.com/dequelabs/axe-test-fixtures/pull/12/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R304

packages/webdriverio/error-handling.md
```diff
@@ -0,0 +1 @@
+# Error Handling
```
WilcoFiers, line 1: Where's the rest of it?

WilcoFiers (changes requested):
Why didn't you adopt this part of the proposal:
https://github.com/dequelabs/axe-api-team-proposals/pull/204/files#diff-96a02d97446db7824b84e655002469cbb37be2a86b4446e528b368ee3b2e363eR73

packages/playwright/error-handling.md
```diff
@@ -0,0 +1 @@
+# Error Handling
```
stephenmathieson, line 1: This is a lovely document 😆

packages/playwright/src/index.ts
stephenmathieson, line 155: Nitpick - prefixing the error message with "Error:" is redundant.

```
> throw new Error('Error: blah blah blah\nPlease check out https://......')
Uncaught Error: Error: blah blah blah
Please check out https://......
>
```

This now reads _Uncaught Error: Error_. Perhaps remove the leading "Error:" text, as it's already provided by the JS engine?


```suggestion
        `${error}\n Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md`
```

stephenmathieson, line 155: I do not recall what the spec says here. Does `Error#toString()` _always_ return `error.message`? If not, we might want to be explicit here:

```suggestion
        `Error: ${error.message}\n Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md`
```

stephenmathieson: `error.trace` is not a standard property. Do not use that.

PR #359, opened 2021-09-09
chore(wdio): support shadow frames in devtools protocol

WilcoFiers:
Fixes the issue that iframes in shadow DOM would not get tested when webdriverio was running in shadow DOM.

### DIFF
--- a/packages/webdriverio/src/utils.ts
+++ b/packages/webdriverio/src/utils.ts
@@ -1,8 +1,10 @@
+import { Element } from 'webdriverio';
 import type {
   AxeResults,
-  ElementContext,
+  CrossTreeSelector,
   PartialResult,
-  ContextObject
+  ContextObject,
+  FrameContext
 } from 'axe-core';
 import type {
   BrowserObject,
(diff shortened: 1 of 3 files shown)


packages/webdriverio/src/utils.ts
```diff
@@ -184,3 +185,15 @@ export const axeFinishRun = ({
...
 };
+
+export async function axeShadowSelect(
+  client: BrowserObject,
+  selector: CrossTreeSelector
+): Promise<Element | undefined> {
+  const selectors: string[] = Array.isArray(selector) ? selector : [selector];
+  let elmRef: Element | undefined = await client.$(selectors[0]);
+  for (let i = 1; i < selectors.length && elmRef; i++) {
+    elmRef = await client.$(selectors[i]);
```
straker, line 196: This does not enter shadow dom, it only looks at light dom. Moreover, it grabs the selector from the top-level `client` and does not use the last `elmRef` as a starting point. So this selector`['#shadow-root', 'iframe']` will find the first `iframe` element on the page, not the iframe that is a child of `#shadow-root`.

straker (changes requested):

PR #362, opened 2021-09-09
chore(webdriverio): use wdio to find frames in page

michael-siek:


straker (commented):
LGTM, but @WilcoFiers should approve since I helped pair on it.

packages/webdriverio/README.md
```diff
@@ -153,3 +153,7 @@ const axe = new AxeBuilder({ client }).setLegacyMode();
 const result = await axe.analyze();
 axe.setLegacyMode(false); // Disables legacy mode
 ```
+
+### Shadow DOM support
+
+`@axe-core/webdriverio` is unable to support shadow DOM elements due to lack of shadow DOM support in [webdriverio](https://webdriver.io/) devtools protocol.
```
WilcoFiers, line 159: That's not true at all. What isn't supported is iframes inside shadow DOM.

packages/webdriverio/src/test.ts
```diff
@@ -118,6 +71,10 @@ describe('@axe-core/webdriverio', () => {
       server.close();
     });
     describe('AxeBuilder', () => {
+      it('check to make sure that client is running devtools protocol', () => {
+        assert.isTrue(client.isDevTools);
```
WilcoFiers, line 75: I was under the impression you were going to test both webdriver and devtools. I feel like only testing one is a mistake.

WilcoFiers (dismissed):

nloding (approved):

PR #364, opened 2021-09-14
feat(wdio): Upgrade to support, and use types of v7

WilcoFiers:
Adds support for WDIO v7. This changes the types over to use v7's types. Using axe-core/webdriverio v4.3 with wdio v6 will now see `any` as the argument for `client`.

### DIFF
--- a/packages/webdriverio/package.json
+++ b/packages/webdriverio/package.json
@@ -61,7 +61,7 @@
     "webdriverio": "^7.12.4"
   },
   "peerDependencies": {
-    "webdriverio": "^6.0.0 || ^5.0.0"
+    "webdriverio": "^7.0.0 || ^6.0.0 || ^5.0.0"
   },
   "nyc": {
     "include": [
--- a/packages/webdriverio/package.json
+++ b/packages/webdriverio/package.json
@@ -44,7 +44,7 @@
     "@types/mocha": "^8.2.3",
     "@types/node": "^14.17.9",
     "@types/test-listen": "^1.1.0",
-    "@wdio/sync": "^6.11.0",
+    "@wdio/sync": "^7.12.4",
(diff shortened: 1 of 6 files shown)


packages/webdriverio/src/index.ts
```diff
@@ -21,37 +12,53 @@ import {
...
-    this.client = client;
-    this.axeSource = axeSource || source;
-    this.includes = [];
-    this.excludes = [];
-    this.option = {};
-    this.disableFrameSelectors = [];
+    // Treat everything as Browser<'async'>:
+    // - Anything sync can also run async, since JS can await sync functions
+    // - Ignore MultiRemoteBrowser, which is just Browser with extra props
+    this.client = client as Browser<'async'>;
```
WilcoFiers, line 42: Needed to simplify this a little. Typescript does not like the union type `.execute()` here.

michael-siek (dismissed):

packages/webdriverio/package.json
```diff
@@ -58,7 +58,7 @@
     "ts-node": "^9.1.1",
     "tsc-silent": "^1.2.1",
     "typescript": "^4.3.5",
-    "webdriverio": "^6.12.1"
+    "webdriverio": "^7.12.4"
   },
   "peerDependencies": {
     "webdriverio": "^6.0.0 || ^5.0.0"
```
straker, line 64:
```suggestion
    "webdriverio": "^7.0.0 || ^6.0.0 || ^5.0.0"
```

straker (changes requested):

packages/webdriverio/package.json
WilcoFiers: Oh!!!! :/ I swear I made that change. Must have messed it up with an npm install.

straker (approved):

Issue #310, opened 2021-07-29
@axe-core/webdriverjs "setting up your environment" link in README.md is broken


straker:
The "setting up your environment" link is broken [on the npm page](https://www.npmjs.com/package/@axe-core/webdriverjs) in that it requires npm login credentials to view the contributing file (is that even the right file to link to?)

PR #388, opened 2021-10-13
[wip]docs: add `CONTRIBUTING.md`

Zidious:
closes issue: https://github.com/dequelabs/axe-core-npm/issues/310

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,0 +1,68 @@
+# Contributing
+
+## Contribution Guidelines
+
+### Code Quality
+
+Although we do not have official coding guidelines, we can and will request you to make changes if we think that your code is sloppy. You can take clues from the existing code base to see what we consider to be reasonable code quality. Please be prepared to make changes that we ask of you even if you might not agree with the request(s).
+
+Pull requests that change the tabs of a file (spacing or changes from spaces to tabs and vice versa) will not be accepted. Please respect the coding style of the files you are changing and adhere to that.
+
+That having been said, we prefer:
+
+1. Tabs over spaces
+2. Single quotes for string literals
+3. Function definitions like `function functionName(arguments) {`
+4. Variable function definitions like `Class.prototype.functionName = function (arguments) {`
+5. Use of 'use strict'
+6. Variables declared at the top of functions
+
+### Documentation and Comments
+
+Functions should contain a preceding comment block with [jsdoc](http://usejsdoc.org/) style documentation of the function. For example:
+
+```
+/**
+ * Runs the Audit; which in turn should call `run` on each rule.
+ * @async
+ * @param  {Context}   context The scope definition/context for analysis (include/exclude)
+ * @param  {Object}    options Options object to pass into rules and/or disable rules or checks
+ * @param  {Function} fn       Callback function to fire when audit is complete
+ */
+```
+
+Classes should contain a jsdoc comment block for each attribute. For example:
+
+```
+/**
+ * Constructor for the result of checks
+ * @param {Object} check CheckResult specification
+ */
+function CheckResult(check) {
+	'use strict';
+
+	/**
+	 * ID of the check.  Unique in the context of a rule.
+	 * @type {String}
+	 */
+	this.id = check.id;
+
+	/**
+	 * Any data passed by Check (by calling `this.data()`)
+	 * @type {Mixed}
+	 */
+	this.data = null;
+
+	/**
+	 * Any node that is related to the Check, specified by calling `this.relatedNodes([HTMLElement...])` inside the Check
+	 * @type {Array}
+	 */
+	this.relatedNodes = [];
+
+	/**
+	 * The return value of the Check's evaluate function
+	 * @type {Mixed}
+	 */
+	this.result = null;
+}
+```
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -1,5 +1,9 @@
 # Contributing
 
+## Contributor License Agreement
+
+In order to contribute, you must accept the [contributor license agreement](https://cla-assistant.io/dequelabs/axe-core-npm) (CLA). Acceptance of this agreement will be checked automatically and pull requests without a CLA cannot be merged.
+
 ## Contribution Guidelines
 
 ### Code Quality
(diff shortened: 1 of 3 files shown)


docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+### Code Quality
+
+Although we do not have official coding guidelines, we can and will request you to make changes if we think that your code is sloppy. You can take clues from the existing code base to see what we consider to be reasonable code quality. Please be prepared to make changes that we ask of you even if you might not agree with the request(s).
+
+Pull requests that change the tabs of a file (spacing or changes from spaces to tabs and vice versa) will not be accepted. Please respect the coding style of the files you are changing and adhere to that.
+
+That having been said, we prefer:
+
+1. Tabs over spaces
+2. Single quotes for string literals
```
stephenmathieson, line 18: Prettier does this for us. Why include it here?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+3. Function definitions like `function functionName(arguments) {`
```
stephenmathieson, line 19: What's wrong with arrow functions? Based on the codebase, we tend to prefer them 🤔

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+4. Variable function definitions like `Class.prototype.functionName = function (arguments) {`
```
stephenmathieson, line 20: Why not:

```js
class MyClass {
  functionName (args) {
    // ...
  }
}
```

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+5. Use of 'use strict'
```
stephenmathieson, line 21: This is implied by using ES modules/TypeScript.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+6. Variables declared at the top of functions
```
stephenmathieson, line 22: Why? With the addition of block scoped variables (`let`, `const`), this rule seems pointless.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+Functions should contain a preceding comment block with [jsdoc](http://usejsdoc.org/) style documentation of the function. For example:
+
+```
+/**
+ * Runs the Audit; which in turn should call `run` on each rule.
+ * @async
+ * @param  {Context}   context The scope definition/context for analysis (include/exclude)
+ * @param  {Object}    options Options object to pass into rules and/or disable rules or checks
+ * @param  {Function} fn       Callback function to fire when audit is complete
+ */
```
stephenmathieson, line 35: We should be pushing for TypeScript, making the type redundant here.

```suggestion
/**
 * Runs the Audit; which in turn should call `run` on each rule.
 * @async
 * @param context The scope definition/context for analysis (include/exclude)
 * @param options Options object to pass into rules and/or disable rules or checks
 * @param fn Callback function to fire when audit is complete
 */
```

---

What is `@async`?

```suggestion
/**
 * Runs the Audit; which in turn should call `run` on each rule.
 *
 * @param  {Context}   context The scope definition/context for analysis (include/exclude)
 * @param  {Object}    options Options object to pass into rules and/or disable rules or checks
 * @param  {Function} fn       Callback function to fire when audit is complete
 */
```

---

The spacing is off.

```suggestion
/**
 * Runs the Audit; which in turn should call `run` on each rule.
 * @async
 * @param  {Context}   context The scope definition/context for analysis (include/exclude)
 * @param  {Object}     options Options object to pass into rules and/or disable rules or checks
 * @param  {Function} fn       Callback function to fire when audit is complete
 */
```

---

All suggestions merged:

```suggestion
/**
 * Runs the Audit; which in turn should call `run` on each rule.
 *
 * @param context The scope definition/context for analysis (include/exclude)
 * @param options Options object to pass into rules and/or disable rules or checks
 * @param fn Callback function to fire when audit is complete
 */
```

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+```
+
+Classes should contain a jsdoc comment block for each attribute. For example:
```
stephenmathieson, line 38:
```suggestion
Classes should contain a JSDoc comment block for each attribute. For example:
```

Also what is "attribute" here?

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+
+```
+/**
+ * Constructor for the result of checks
+ * @param {Object} check CheckResult specification
```
stephenmathieson, line 43:
```suggestion
 *
 * @param check CheckResult specification
```

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+ */
+function CheckResult(check) {
+	'use strict';
```
stephenmathieson, line 46: I am _strongly_ against opting in to strict mode in function scopes. Either the entire module/file should be in strict mode or none of it.

Again, using ESM makes this moot.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+
+	/**
+	 * ID of the check.  Unique in the context of a rule.
+	 * @type {String}
+	 */
+	this.id = check.id;
+
+	/**
+	 * Any data passed by Check (by calling `this.data()`)
+	 * @type {Mixed}
```
stephenmathieson, line 56: `Mixed` is not a type.

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	this.data = null;
+
+	/**
+	 * Any node that is related to the Check, specified by calling `this.relatedNodes([HTMLElement...])` inside the Check
+	 * @type {Array}
```
stephenmathieson, line 62: Array of what?

```suggestion
	 * @type {Array<TheType>}
```

Or better yet:

```suggestion
	 * @type {TheType[]}
```

docs/CONTRIBUTING.md
```diff
@@ -0,0 +1,72 @@
...
+	 */
+	this.relatedNodes = [];
+
+	/**
+	 * The return value of the Check's evaluate function
+	 * @type {Mixed}
```
stephenmathieson, line 68: Not a valid type.

PR #402, opened 2021-11-19
fix(webdriverjs): prevent selnium undefined -> null transformation

AdnoC:


WilcoFiers (changes requested):
tests

packages/webdriverjs/src/browser.ts
```diff
@@ -49,7 +49,7 @@ export function axeRunPartial(
       var callback = arguments[arguments.length - 1];
       var context = ${JSON.stringify(context)} || document;
       var options = ${JSON.stringify(options)} || {};
-      window.axe.runPartial(context, options).then(callback);
+      window.axe.runPartial(context, options).then(res => JSON.parse(JSON.stringify(res))).then(callback);
```
WilcoFiers, line 52: How does this solve the issue?

AdnoC: The issue is that there are places where a key is set to `undefined` rather than deleting it. Selenium finds these and converts them to `null`. JSON roundtrip removed the keys

WilcoFiers: Why not just return the string, so you don't have to serialise twice

WilcoFiers (changes requested):
This needs a test

Issue #332, opened 2021-08-16
axe-core/webdriverjs and webdriverio should support complex include / exclude


WilcoFiers:
This should be possible in webdriverjs / webdriverio

```js
new AxeBuilder()
  .include(['#myIframe', 'html'])
  .exclude(['#myIframe', 'main'])
```

This is possible in axe-core/puppeteer, but not in webdriverjs / webdriverio.

When we fix this, we should add better tests. Test coverage for include/exclude methods is kinda poor.

padmavemulapati:
Validated during axe-core-npm 4.4.2 regression, on passing `include` and `exclude` options together with and without `iframes` , working as expected.

PR #404, opened 2021-11-29
fix(webdriverJS): include/exclude chaining and iframe selectors

Zidious:
Added support for method chaining on include and include as well as iframe CSS selectors. 

Refactored the current include/exclude tests to check if the results include / exclude the correct class. Used a function from [Playwright](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/tests/axe-playwright.spec.ts#L377) that returns the node targets within a string array.

Examples: 

```js
// include and exclude
 const builder = new AxeBuilder(driver)
        .include('.include')
        .exclude('.exclude');
```
```js
// chaining include
const builder = new AxeBuilder(driver)
        .include('.include')
        .include('.include2');
```
```js
// chaining exclude
const builder = new AxeBuilder(driver)
        .exclude('.exclude')
        .exclude('.exclude2');
```
```js
// chaining include and exclude
const builder = new AxeBuilder(driver)
 		.include('.include')
        .include('.include2')
        .exclude('.exclude')
        .exclude('.exclude2');
```
```js
// iframe selectors
const builder = new AxeBuilder(driver)
        .include(['#ifr-one', 'html'])
        .exclude(['#ifr-one', 'main']);
```

Closes Issue: https://github.com/dequelabs/axe-core-npm/issues/332 (WDJS part)

### DIFF
--- a/packages/webdriverjs/src/index.ts
+++ b/packages/webdriverjs/src/index.ts
@@ -1,7 +1,12 @@
 import { WebDriver } from 'selenium-webdriver';
 import { RunOptions, Spec, AxeResults, ContextObject } from 'axe-core';
 import { source } from 'axe-core';
-import { CallbackFunction, BuilderOptions, PartialResults } from './types';
+import {
+  CallbackFunction,
+  BuilderOptions,
+  PartialResults,
+  Selector
+} from './types';
 import { normalizeContext } from './utils/index';
 import AxeInjector from './axe-injector';
 import {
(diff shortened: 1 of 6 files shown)


packages/webdriverjs/src/index.ts
```diff
@@ -16,8 +16,8 @@ import * as assert from 'assert';
 class AxeBuilder {
   private driver: WebDriver;
   private axeSource: string;
-  private includes: string[];
-  private excludes: string[];
+  private includes: string[][];
+  private excludes: string[][];
```
WilcoFiers, line 20: This is confusing. Can you use the actual types axe-core uses for them?

AdnoC (approved):

PR #408, opened 2021-11-30
chore(release-tests): new package to test releases went out correctly

WilcoFiers:


WilcoFiers:
Both; Test that the release we just did went out, and that it can be installed and run without errors.

nloding:
OK, I think I was expecting to see a test along these lines:

```ts
const playwright = require('playwright');
const { default: AxeBuilder } = require('@axe-core/playwright');
const { version } = require('@axe-core/playwright/package.json');
const { version: releaseVersion } = require('../package.json')

describe(`@axe-core/playwright v${version}`, () => {
  it('has the right version', async () => {
    assertEqual(version, releaseVersion)
  });
});
```

After the release runs, the version should be bumped on this package, and if we keep them in sync ...

Would this not work?

WilcoFiers:
No, the version in the package.json root is just 1.0.0. But I suppose we could pull out each package.json file from packages and compare it to that. That'd do it.

nloding (approved):
I like it!

Issue #398, opened 2021-10-27
allow playwright to accept multi-array selectors 


Zidious:
Right now we have the ability to chain multiple includes/excludes:
```js
const builder = new AxeBuilder()
  .exclude('.exclude1')
  .exclude('.exclude2')
  
// or 
  
const builder = new AxeBuilder()
  .exclude(['.exclude1'])
  .exclude(['.exclude2'])
```
But not in the same array: 
```js
const builder = new AxeBuilder()
  .exclude(['.exclude1', '.exclude2'])
```
The above does not properly exclude classes `.exclude1` and `.exclude2`. 

Resource: [axe-core include/exclude](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#include-exclude-object)

edit: my apologies, the above include/exclude example is not correct. the correct layout for include/exclude for classes would look something like: 

```js
const builder = new AxeBuilder()
  .exclude([ ['.exclude1'], ['.exclude2'] ])
  ```

See bullet point 3 - Resource: [axe-core include/exclude](https://github.com/dequelabs/axe-core/blob/develop/doc/API.md#include-exclude-object)

Ref: https://github.com/dequelabs/axe-core-npm/pull/391

Bracciata:
This should be fixed!

Zidious:
See: https://github.com/dequelabs/axe-core-npm/pull/413#issuecomment-1007520245

PR #413, opened 2021-12-16
Fix uploading of include and exclude from multi string array #398

Bracciata:


michael-siek:
Thank you very much for your contribution!

Zidious (changes requested):
Hey @Bracciata! 

Thanks for contributing, unfortunately the issue I created was not quite correct. The current include / exclude selectors I put in the issue [ticket](https://github.com/dequelabs/axe-core-npm/issues/398) show iframe selectors and not the correct layout for include/excluding whom parent include said classes as shown in in the axe-core example:

> Include the whole document except any structures whose parent contains the class `exclude1` or `exclude2`. the structure would need to look like: 

```js
const builder = new AxeBuilder()
  .exclude([ ['.exclude1'], ['.exclude2'] ])
  ```

File: packages/playwright/tests/axe-playwright.spec.ts [hunk 1]
```diff
@@ -459,48 +459,38 @@ describe('@axe-core/playwright', () => {
...
-        '.selector-three'
-      ]);
+      const results = await new AxeBuilder({ page })
+        .include(['.include', '.include2'])
+        .analyze();
 
-      const { include: actual } = (await results.analyze()) as any;
+      const flattenTarget = flatPassesTargets(results);
 
-      assert.deepEqual(actual[0], expected);
```
Zidious, line 480: We would need to keep this test, it checks that the selectors are getting passed to axe-core

packages/playwright/tests/axe-playwright.spec.ts
```diff
@@ -459,48 +459,38 @@ describe('@axe-core/playwright', () => {
...
+    });
+    it('with chaining include/exclude using an array of strings', async () => {
+      await page.goto(`${addr}/context.html`);
+      const results = await new AxeBuilder({ page })
+        .include(['.include', '.include2'])
+        .exclude(['.exclude', '.exclude2'])
+        .analyze();
+      const flattenTarget = flatPassesTargets(results);
 
-      assert.deepEqual(actual[0], expected);
```
Zidious, line 503: We would need to keep this exclude test, it checks that the exclude selectors are getting passed to axe-core

Bracciata:
@Zidious What are your thoughts on fully flattening the array then appending?

Zidious:
> @Zidious What are your thoughts on fully flattening the array then appending?

@Bracciata Thanks for getting back to me! I believe that will not work as its how it gets passed to axe-core. Currently, we allow chaining of include/exclude which produces the correct array format, example: 

```js
const builder = new AxeBuilder()
  .exclude('.exclude1')
  .exclude('.exclude2')
  
  // When produces this when passed to axe-core: { exclude: [ [ '.exclude' ], [ '.exclude2' ] ] }
```
So we would need to mimic the same behaviour, leave it with me, I will need to discuss it further with the team and I will get back to you ASAP.

Zidious:
@Bracciata Sorry to keep you waiting, had a discussion with the team due to how the include / exclude array is passed to axe-core it would require a significant amount of changes. We have decided to keep the current include / exclude chaining for classes (See below): 

```js
const builder = new AxeBuilder()
  .exclude('.exclude1')
  .exclude('.exclude2')
  ```

Issue #370, opened 2021-09-17
axe cli apparently not respecting the --chromedriver-path argument


macbookandrew:
Product: cli

Expectation: I expect that using the `--chromedriver-path` to specify a path should work

Actual: It apparently ignores it

Example:

```shell
❯ axe https://example.com --chromedriver-path="/usr/local/Caskroom/chromedriver/93.0.4577.63/chromedriver" --show-errors
Running axe-core 4.3.3 in chrome-headless
Error: "session not created: This version of ChromeDriver only supports Chrome version 90\nCurrent browser version is 93.0.4577.82 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
 $s SessionNotCreatedError: session not created: This version of ChromeDriver only supports Chrome version 90
Current browser version is 93.0.4577.82 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome
    at Object.throwDecodedError (/usr/local/lib/node_modules/@axe-core/cli/node_modules/selenium-webdriver/lib/error.js:517:15)
    at parseHttpResponse (/usr/local/lib/node_modules/@axe-core/cli/node_modules/selenium-webdriver/lib/http.js:643:13)
    at Executor.execute (/usr/local/lib/node_modules/@axe-core/cli/node_modules/selenium-webdriver/lib/http.js:569:28)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
Please report the problem to: https://github.com/dequelabs/axe-core-npm/issues/

# To verify chromedriver version:
❯ /usr/local/Caskroom/chromedriver/93.0.4577.63/chromedriver --version
ChromeDriver 93.0.4577.63 (ff5c0da2ec0adeaed5550e6c7e98417dac77d98a-refs/branch-heads/4577@{#1135})
```

Motivation: I’m just trying to use the CLI 😂 

<pre><code>
axe-core version: 4.3.3
@axe-core/cli: 4.2.2

- Node version: 16.6.1
- Platform: Mac
</code></pre>

macbookandrew:
BTW I am getting the same error when I run without the `--chromedriver-path` flag:

```shell
❯ axe https://example.com --show-errors
Running axe-core 4.3.3 in chrome-headless
Error: "session not created: This version of ChromeDriver only supports Chrome version 90\nCurrent browser version is 93.0.4577.82 with binary path /Applications/Google Chrome.app/Contents/MacOS/Google Chrome"
[...]

PR #416, opened 2022-01-05
fix(cli): pass chromedriver-path arg to webdriver

Zidious:
We were not passing the `--chromedriver-path` arg value to webdriver. 

Within `webdriver.ts` `config.chromedriverPath` would produce undefined even if `--chromedriver-path` was set which would result in using the default chrome driver and not the one set by the user.

image
  
If we add `chromedriverPath` to the destructured args and a user provides an invalid path we do error correctly throw the error as well as use the given Chrome driver: 

Invalid Path:

image

Valid Path: 

image

Note:  `--chromedriver-path` does not seem to support backwards compatibility in terms of Chrome Driver, it depends on your personal Chrome version. Example, if you're using V96 you cannot use V95 chrome driver, otherwise you will get a session error (as shown in the issue referenced).

Closes issue: https://github.com/dequelabs/axe-core-npm/issues/370

### DIFF
--- a/packages/cli/src/bin/cli.test.ts
+++ b/packages/cli/src/bin/cli.test.ts
@@ -315,4 +315,18 @@ describe('cli', () => {
       }
     });
   });
+
+  describe('--chromedriver-path', () => {
+    it('should throw error if path does not exist', async () => {
+      const result = await runCLI(
+        `file://${SIMPLE_HTML_FILE}`,
+        '--chromedriver-path="someinvalidpath"',
+        '--show-errors'
+      );
+      assert.include(
+        result.stderr,
+        'The specified executable path does not exist'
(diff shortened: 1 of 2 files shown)


michael-siek:
We _should_ add a test for this.

Zidious:
> We _should_ add a test for this.


I've added a test for checking when the driver [path does not exist](https://github.com/dequelabs/axe-core-npm/pull/416/files#diff-ba589c4bddd3dc1debf232e58e3056c8f606bada79767aa85a57b1d777874731R319-R331). I am not sure if we can test if the driver path does exist and run CLI. would require us to have a Chrome driver executable somewhere.

WilcoFiers (approved):

Issue #200, opened 2021-02-26
Examples section was removed for webdriverjs package


jeeyyy:
Currently published package of `@axe-core/webdriverjs` has examples section which will be super helpful if the links worked :)
> See https://www.npmjs.com/package/@axe-core/webdriverjs

The new re-write to TS seems to have lost the examples section, this should come back.
> See - https://github.com/dequelabs/axe-core-npm/tree/develop/packages/webdriverjs

michael-siek:
Adding for context: https://github.com/dequelabs/axe-webdriverjs/blob/develop/README.md

PR #420, opened 2022-01-10
docs(webdriverjs): add example using axe-core webdriverjs

Zidious:
closes issue: https://github.com/dequelabs/axe-core-npm/issues/200


packages/webdriverjs/tests/example/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "@types/node": "^14.17.32",
+    "@types/selenium-webdriver": "^4.0.16",
+    "chai": "^4.3.0",
+    "mocha": "^8.4.0",
+    "selenium-webdriver": "^4.0.0",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.4.4"
+  },
+  "dependencies": {
+    "@axe-core/webdriverjs": "^4.3.2"
```
WilcoFiers, line 20: How are we going to keep this up to date? Should we link this with lerna?

packages/webdriverjs/tests/example/package.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "axe-core-npm-webdriverjs-example",
+  "private": "true",
+  "scripts": {
+    "test": "mocha --timeout 60000 -r ts-node/register 'webdriverjs-example.ts'"
```
WilcoFiers, line 5: I think we should run this in CI.

WilcoFiers (dismissed):

packages/webdriverjs/tests/example/webdriverjs-example.ts
```diff
@@ -0,0 +1,85 @@
...
+      .build();
+  });
+
+  afterEach(() => {
+    driver.close();
+  });
+
+  describe('Deqeue Homepage', () => {
+    it('should run analysis', async () => {
+      await driver.get('https://deque.com/');
```
stephenmathieson, line 22: If this is going to be run in CI, we should not be relying on `deque.com` being up. If the marketing website goes down (maybe they're doing an upgrade, or perhaps Wordpress has exploded), our tests will fail. This will prevent us from hot-fixing critical bugs quickly.

I **strongly** recommend running a local server and testing "fixtures" rather than a live website (which this team does not control).

packages/webdriverjs/tests/example/webdriverjs-example.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    it('should find violations', async () => {
+      await driver.get('https://deque.com/');
+
+      const results = await new AxeBuilder(driver).analyze();
+
+      assert.lengthOf(results.violations, 2);
+      assert.equal(results.violations[0].id, 'landmark-no-duplicate-banner');
+      assert.equal(results.violations[1].id, 'landmark-unique');
```
stephenmathieson, line 39: This seems brittle. Is the order violations are returned from axe-core part of its public API? It seems like this could change in a patch release.

We ought to make this a little more robust.

```suggestion
      assert.isDefined(results.violations.find(v => v.id === 'landmark-no-duplicate-banner'))
      assert.isDefined(results.violations.find(v => v.id === 'landmark-unique'))
```

packages/webdriverjs/tests/example/webdriverjs-example.ts
```diff
@@ -0,0 +1,85 @@
...
+    });
+
+    it('should exclude CSS selector', async () => {
+      await driver.get('https://deque.com/');
+
+      const results = await new AxeBuilder(driver)
+        .exclude('#masthead')
+        .analyze();
+
+      assert.lengthOf(results.violations, 1);
```
stephenmathieson, line 49: This also seems brittle. What if axe-core adds a new rule? What if a new a11y issue appears on deque.com?

stephenmathieson (dismissed):
If this is going to run on CI, some changes are necessary.

Zidious:
CI Workflow: https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm/1005/workflows/5efe4d99-e300-41fb-a838-c2697002e450

CI Webdriverjs-example tests: https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm/1005/workflows/5efe4d99-e300-41fb-a838-c2697002e450/jobs/7897?invite=true#step-103-9

packages/webdriverjs/tests/example/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "axe-core-npm-webdriverjs-example",
+  "name": "@axe-core-npm/webdriverjs-example",
   "private": "true",
```
WilcoFiers, line 3: This might be important: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#private

```suggestion
  "private": true,
```

packages/webdriverjs/tests/example/package.json
```diff
@@ -11,12 +11,11 @@
...
     "chai": "^4.3.0",
+    "express": "^4.17.2",
     "mocha": "^8.4.0",
     "selenium-webdriver": "^4.0.0",
+    "test-listen": "^1.1.0",
     "ts-node": "^9.1.1",
     "typescript": "^4.4.4"
-  },
-  "dependencies": {
-    "@axe-core/webdriverjs": "^4.3.2"
```
WilcoFiers: You'll still need this as a dependency. Maybe just use `*` as its version.

packages/webdriverjs/tests/example/webdriverjs-example.ts
```diff
@@ -1,12 +1,23 @@
 import { assert } from 'chai';
 import { Builder, WebDriver } from 'selenium-webdriver';
+import AxeBuilder from '../../src';
```
WilcoFiers, line 3: You can still use Lerna to link them up I think. Makes for a better example that way.

packages/webdriverjs/tests/example/webdriverjs-example.ts
```diff
@@ -1,12 +1,23 @@
...
+describe('@axe-core/webdriverjs example', () => {
   let driver: WebDriver;
+  let server: Server;
+  let addr: string;
+
+  beforeEach(async () => {
+    const app = express();
+    app.use(express.static(path.resolve(__dirname, '..', 'fixtures')));
+    server = createServer(app);
+    addr = await testListen(server);
```
WilcoFiers, line 19: Doesn't seem like this should run each test. Better to just use a `before` here.

WilcoFiers (changes requested):

lerna.json
```diff
@@ -1,4 +1,9 @@
 {
-  "packages": ["packages/*", "packages/react/examples/*", "axe_core_test"],
+  "packages": [
+    "packages/*",
+    "packages/react/examples/*",
+    "packages/webdriverjs/tests/*",
```
stephenmathieson, line 5: It looks like the are directories other than the examples here. This should likely be:

```suggestion
    "packages/webdriverjs/tests/example",
```

packages/webdriverjs/README.md
```diff
@@ -162,3 +162,13 @@ const axe = new AxeBuilder(driver).setLegacyMode();
 const result = await axe.analyze();
 axe.setLegacyMode(false); // Disables legacy mode
 ```
+
+### WebdriverJS Deque example
```
stephenmathieson, line 166: Why mention "Deque" here?

```suggestion
### WebdriverJS Example
```

Zidious: Forgot to remove after using local server as example instead of deque.com

WilcoFiers (approved):

PR #421, opened 2022-01-10
fix(cli): provide a more robust error message if analysis fails

Zidious:
The use of `--include` and providing an non-existent or invalid CSS selector would not notify user that was the case: 

image
image

Providing an non-existent or invalid CSS selector will now notify the user of said error: 

image

Closes issue: https://github.com/dequelabs/axe-core-npm/issues/207

### DIFF
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -1,7 +1,6 @@
 import * as WebDriver from 'selenium-webdriver';
 import AxeBuilder from '@axe-core/webdriverjs';
 import { AxeResults } from 'axe-core';
-import { error } from '../lib/utils';
 import { EventResponse, ConfigParams } from '../types';
 
 const testPages = async (
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -1,6 +1,7 @@
 import * as WebDriver from 'selenium-webdriver';
 import AxeBuilder from '@axe-core/webdriverjs';
 import { AxeResults } from 'axe-core';
+import { error } from '../lib/utils';
 import { EventResponse, ConfigParams } from '../types';
 
 const testPages = async (
@@ -91,6 +92,7 @@ const testPages = async (
 
           /* istanbul ignore if */
           if (err) {
+            console.error(error(err));
             return reject(err);
           }
 
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -92,8 +92,7 @@ const testPages = async (
 
           /* istanbul ignore if */
           if (err) {
-            console.error(error(err));
-            process.exit(1);
+            return reject(err);
           }
 
           // Notify about the update
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -93,7 +93,7 @@ const testPages = async (
           /* istanbul ignore if */
           if (err) {
             console.error(error(err));
-            return reject(err);
+            process.exit(1);
           }
 
           // Notify about the update
--- a/packages/cli/src/bin/cli.test.ts
+++ b/packages/cli/src/bin/cli.test.ts
@@ -185,7 +185,8 @@ describe('cli', () => {
       const result = await runCLI(
         `file://${SIMPLE_HTML_FILE}`,
         '--include',
-        '#hazaar'
+        '#hazaar',
+        '--show-errors'
       );
 
       assert.include(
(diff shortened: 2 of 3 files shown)


packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -91,7 +92,8 @@ const testPages = async (
 
           /* istanbul ignore if */
           if (err) {
-            return reject(err);
+            console.error(error(err));
+            process.exit(1);
```
stephenmathieson, line 96: Why doesn't `reject(err)` allow the caller to handle the error on their own? I don't understand how/why this fixes anything.

Zidious: `reject(err)` doesn't provide a useful error message to the user: 

image

Including `--show-errors` flag does not either: 

image

stephenmathieson: That seems like the problem is how we're handing rejections then.

stephenmathieson: It looks like the issue is here:

https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/cli/src/bin/index.ts#L153

What is `%j \n $s` supposed to do?

Zidious: I think the error is here https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/cli/src/lib/axe-test-urls.ts#L87 we use `err: string` so we don't have access to `e.message` or `e.stack` to do the the string formatting you mentioned

stephenmathieson: I see. Yes, we should _only_ be rejecting with `Error`s, not other things. Rejecting/`throw`ing anything but an instance of an `Error` is considered a bad practice and should be avoided. [ESLint protects us from this sort of thing with `throw`](https://eslint.org/docs/rules/no-throw-literal), but doesn't do the same for Promise rejections. We should pass the entire error back, not just a string.

I didn't know about `%j`, thanks for that. FWIW the `Error` class does not have a `.toJSON()`, so I'm pretty sure that's not doing anything.

stephenmathieson: If the types on the line you linked are correct (`err: string`), then we should do something like:

```ts
        axe.analyze((err: Error | string | null, results: AxeResults) => {
          if (config.timer) {
            events?.endTimer('axe-core execution time');
          }

          /* istanbul ignore if */
          if (err) {
            const error = typeof err === 'string' ? new Error(err) : err
            return reject(error);
          }
```

stephenmathieson: It looks like that type is wrong, as we [pass back the whole `Error` (unless in legacy/callback mode)](https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/webdriverjs/src/index.ts#L134).

Zidious: Ya, looking at it a bit more, the [type for the callback function](https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/webdriverjs/src/types.ts#L20) should be `Error | null` and not `string | null`. So we can access `e.message` etc. However, your work around does allow the user to _handle_ the error with `--show-errors`

image

stephenmathieson: Yes, the type for that function is wrong. IMO it should be changed/corrected.

However, if we want a _very simple fix_ for the issue at hand, we could just fix the `console.error()` call in `src/bin/index.ts`:

```js
    if (!showErrors) {
      console.error(error('An error occurred while testing this page.'));
    } else {
      console.error(error('Error: %s'), e);
    }
```

packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -1,6 +1,7 @@
 import * as WebDriver from 'selenium-webdriver';
 import AxeBuilder from '@axe-core/webdriverjs';
 import { AxeResults } from 'axe-core';
+import { error } from '../lib/utils';
```
stephenmathieson, line 4:
```suggestion
```

stephenmathieson (approved):

packages/cli/src/lib/axe-test-urls.ts
Zidious: resolved!

packages/cli/src/lib/axe-test-urls.ts
Zidious: For reference issue has been raised: https://github.com/dequelabs/axe-core-npm/issues/422

stephenmathieson (approved):

WilcoFiers (changes requested):
While I support the change, I don't see how the title describes what you changed, or how this closes the issue.

Issue #308, opened 2021-07-25
axe-cli: "An error occurred while testing this page"


DanielTripp:
Product:  cli 

Expectation:  To print a list of accessibility violations to stdout.   

Actual:  No list of accessibility violations, or any evidence of a scan, appears on stdout.  Instead, this appears:

<pre><code>
DevTools listening on ws://127.0.0.1:54164/devtools/browser/8e985c0a-b6d8-46ba-aabd-592db8ed88a2
An error occurred while testing this page.
Please report the problem to: https://github.com/dequelabs/axe-core-npm/issues/
</code></pre>

Motivation:  I think that my "Expectation" is a design goal of axe-cli.  Of course, I'm new to this, so I could be wrong.   

<pre><code>
axe-core version: 4.2.0, I think.  (That's what it says in my file C:\Users\dt\axe-cli-july-25-b\axe-core-npm\node_modules\axe-core\axe.js.  First line: "/*! axe v4.2.0".) 
@axe-core/cli: 4.2.2, I think.  (That's what it says in my file C:\Users\dt\axe-cli-july-25-b\axe-core-npm\packages\cli\package-lock.json: "version": "4.2.2".)

- Node version: v14.17.3
- Platform:   Windows  
</code></pre>

I've attached my steps to reproduce as a copy-and-paste of my PowerShell console.
[axe-cli-problem-july-25.txt](https://github.com/dequelabs/axe-core-npm/files/6874257/axe-cli-problem-july-25.txt)

I don't know whether this is a bug or not.  But I sure would benefit from some guidance.   All suggestions welcome.

AlexTolok:
I have same issue with axe cli on Linux inside circleCI
node v14.10.0 (npm v7.20.1)
Running axe-core 4.3.1 in chrome-headless
I have run script against different pages and it always fails (previously it worked fine with same versions and same web pages)  it looks like problem appeared with latest update (as I am installing latest version each time)

AlexTolok:
@DanielTripp if to use older version of chrome it works for me so I think issue appears only with latest chrome (axe cli works fine on chrome v 91.0.4472.164-1 but fails on 92.0.4515.107-1) also it passes if to change browser to firefox 
Try:
`axe   www.deque.com --tags wcag2a --browser firefox --dir ~/axeReports`

DanielTripp:
[...]

PR #423, opened 2022-01-10
fix(webdriverjs): Reject with actual `Error`s (not strings)

### LINKED ISSUE #422: fix typing in WebdriverJS to improve error handling in CLI
For reference: https://github.com/dequelabs/axe-core-npm/pull/421

The type in webdriverjs, [callback function](https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/webdriverjs/src/types.ts#L20) used in [analyze](https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/webdriverjs/src/index.ts#L122) needs to be of type `Error | null` instead of `string | null` as when used in [CLI](https://github.com/dequelabs/axe-core-npm/blob/v4.3.2/packages/cli/src/lib/axe-test-urls.ts#L87), we are unable to properly handle the error as it is of type string and not Error therefore we do not have access to Error's functions i.e. `e.message or e.stack`. Example is when a user uses flag `--include` and provided an invalid / non-existent CSS selector when [`--show-errors` is provided to get _why_ it has failed we are presented with `undefined`](https://github.com/dequelabs/axe-core-npm/pull/421#discussion_r781495249)

stephenmathieson:
This patch updates the error handling in `@axe-core/webdriverjs`, having it return an actual `Error` rather than a string. Rejecting/returning a string (eg `error.message`) leads to tons of bugs/confusion. Providing the `Error` to the caller enables a much richer debugging experience.

Closes #422
Ref #421
Ref #387
Ref #308 (somewhat)
Ref #207

### DIFF
--- a/packages/cli/src/bin/index.ts
+++ b/packages/cli/src/bin/index.ts
@@ -150,7 +150,7 @@ const cli = async (
     if (!showErrors) {
       console.error(error('An error occurred while testing this page.'));
     } else {
-      console.error(error('Error: %s \n%s'), e.message, e.stack);
+      console.error(error('Error: %s\n%s'), e.message, e.stack);
     }
 
     console.error(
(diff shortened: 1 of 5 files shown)


Zidious (approved):
LGTM

Issue #415, opened 2022-01-03
react: not triggering logs for navigation changes of react-router v5


straker:
See https://stackoverflow.com/questions/70481900/how-do-i-get-axe-core-react-to-check-re-renders-of-my-app.

Also see the comments of https://dev.to/stories_of_ren/taking-the-axe-to-inaccessibility-4dkj

PR #437, opened 2022-02-01
@axe-core/react: fallback on _reactInternals

seanparmelee:
I was troubleshooting an issue where subsequent renders were not being checked for issues. Turns out the issue is due to the `_reactInternalFiber` field being renamed to `_reactInternals` in v17: https://github.com/facebook/react/pull/18377.

image

This PR adds a fallback for `_reactInternals`. You can see this issue in action if you bump the shadow-dom example to React 17, run it, and then click on one of the services. You will not see the additional log statement in the console after clicking the service. I expect this to also fix https://github.com/dequelabs/axe-core-npm/issues/415.

Let me know if/how you'd like tests for this change. I'm guessing you'd probably prefer the shadow-dom example to remain at React 16, but maybe we could add another integration test that uses the NextJS example, which is already using React 17?

### DIFF
--- a/packages/react/index.ts
+++ b/packages/react/index.ts
@@ -142,9 +142,11 @@ function logFailureMessage(
 ): void {
   // this exists on axe but we don't export it as part of the typescript
   // namespace, so just let me use it as I need
-  const message: string = ((axeCore as unknown) as AxeWithAudit)._audit.data.failureSummaries[
-    key
-  ].failureMessage(node[key].map(check => check.message || ''));
+  const message: string = (
+    axeCore as unknown as AxeWithAudit
+  )._audit.data.failureSummaries[key].failureMessage(
+    node[key].map(check => check.message || '')
+  );
 
   console.error(message);
 }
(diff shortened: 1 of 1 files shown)


packages/react/index.ts
```diff
@@ -142,9 +142,11 @@ function logFailureMessage(
 ): void {
   // this exists on axe but we don't export it as part of the typescript
   // namespace, so just let me use it as I need
-  const message: string = ((axeCore as unknown) as AxeWithAudit)._audit.data.failureSummaries[
-    key
-  ].failureMessage(node[key].map(check => check.message || ''));
+  const message: string = (
```
seanparmelee, line 145: sorry for this noise; these changes are from the formatter

packages/react/index.ts
```diff
@@ -286,6 +289,8 @@ function addComponent(component: any): void {
   const reactInstanceDebugID = reactInstance._debugID;
   const reactFiberInstance = component._reactInternalFiber || {};
   const reactFiberInstanceDebugID = reactFiberInstance._debugID;
+  const reactInternals = component._reactInternals || {};
```
seanparmelee, line 292: the rest of this code is the fix

straker:
An update on this. We've been able to confirm this fixes the problem, but are trying to figure out the best way to add a test so we can catch something like this in the future.

Zidious:
Hey @seanparmelee, 

We will need to update our entire test suite, so no test required at the moment. We have confirmed your fix works as intended. 

Thank you for contributing

pierceray:
The security review check has been running for days, what is the normal timeline for that succeeding?

Zidious:
We have had to cherry-pick this PR to get CI running. Will be merged shortly. 

Reference: https://github.com/dequelabs/axe-core-npm/pull/437

seanparmelee:
Thanks @Zidious, so it sounds like we can close this one out then?

seanparmelee:
Also, while I was investigating all of this, I noticed that the `_debugID` field has since been removed in React (guessing that's slated for React 18) and logged https://github.com/dequelabs/axe-core-npm/issues/438 to bring this to your attention.

seanparmelee:
Resolved in #455

Issue #452, opened 2022-02-11
update chromedriver to V98 on all core-npm packages


Zidious:
Chrome 98 was released recently, ChromeDriver needs to be updated to match in our packages need to be updated accordingly. 

See: 
https://github.com/dequelabs/axe-core-npm/issues/401
https://github.com/dequelabs/axe-core-npm/issues/450

Zidious:
Update to always use `latest` ChromeDriver on all packages: 

- [ ] CLI
- [ ] Playwright
- [ ] Puppeteer
- [ ] React
- [ ] WebdriverIO
- [ ] WebdriverJS

padmavemulapati:
Validated as part of `axe-core-npm` QA, 
Chromedriver was always installing as a dependency, but now it's using `latest` instead of us pinning a `chromedriver` version. 
For Example:
Prior to change:
if a user has Chrome V97, our package was pinned to V95 chrome driver, this would break CLI
New change:
If a user install CLI it will install latest chromedriver (say V99) their browser needs to be fully up to date

PR #454, opened 2022-02-11
chore: update packages to use  version of ChromeDriver

Zidious:
Closes issue: https://github.com/dequelabs/axe-core-npm/issues/452

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1879,8 +1879,8 @@
       "dev": true
     },
     "axe-test-fixtures": {
-      "version": "github:dequelabs/axe-test-fixtures#df007093b42d3dcdd803b1a94565053d5383871a",
-      "from": "github:dequelabs/axe-test-fixtures#v1"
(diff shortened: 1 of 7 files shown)


.circleci/config.yml
```diff
@@ -127,6 +127,7 @@ jobs:
     steps:
       - checkout
       - restore_dependency_cache
+      - run: npx browser-driver-manager@1.0.4 install chrome chromedriver
```
straker, line 130: Since you're doing this for every package, probably best to install it as a dependency at the root so you don't have to waste time reinstalling it for every job.

straker (changes requested):

.circleci/config.yml
Zidious: Chrome version is at V96 when I ran it on SSH, we would need to store Chrome in a cache or find a docker image that updates the chrome versions?

michael-siek (approved):

Issue #449, opened 2022-02-10
`@axe-core/react` rewrite test suite


Zidious:
The current way the tests are setup in our React repo is not working as intended. 

The current violations in our tests that we test against are coming from the [`example.js`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/react/examples/shadow-dom/src/example.js#L1-L20) axe config setup, not from the Cypress tests. 

The axe-react snippet below is taken from a Cypress test: 

```ts
axe(React, ReactDOM, 0, {}).then(function () {
	expect(win.console.group).to.be.calledWith(
		'%cNew axe issues',
		'color:#d93251;font-weight:normal;'
	);
	done();
)};
```
source: [`cypress/integration/index-test.ts`](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/react/cypress/integration/index-test.js#L30-L34)

`React` and `ReactDOM` imports are not shared between the Cypress tests and the `example.js` in the shadow-dom example. If you [comment out line 7-20](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/react/examples/shadow-dom/src/example.js#L1-L20) in example.js and run the cypress test, you will only get axe-result violations on page load. It will not trigger all of the subsequent component updates with [React Axe](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/react/index.ts#L273-L277).

Moreover, we get two sets of axe-results, notice `cypress.runner.js` are the axe-results from the example.js file within the shadow-dom example. `index.js` are the axe-results from the Cypress test itself, we are testing agains't the React from the example.js file not the React from our tests:

<img width="1213" alt="image" src="https://user-images.githubusercontent.com/41127686/153511747-a1bb0c79-adc8-4ef8-91b4-23634423e3b6.png">

We will need to rewrite the test suite to use the single React import for react-axe and test agains't those results.

PR #466, opened 2022-02-16
tests(react): add proper unit tests

straker:
Add proper unit testing to @axe-core/react. This allows us to better maintain the repo and see changes that would break us before they happen. For example, here is react@17 with [the recent fix](https://github.com/dequelabs/axe-core-npm/pull/455) commented out.

3 test failures on the tests that test component update changes

And here's react@16

all tests passing

Also removed Cypress tests as they were no longer needed and covered by the unit tests.

_NOTE:_ I was unable to figure out how to get jest/enzyme to work with typescript and load the `index.ts` file directly, so just loaded the file from `dist`.

Closes issue: #449


Zidious (approved):
LGTM

PR #481, opened 2022-03-15
fix(react): revert export to not use default

straker:
Fixing the accidental breaking change when we switched to use `export default` [instead of `export =`](https://github.com/dequelabs/axe-core-npm/pull/466/files#diff-037b868307272ef21e46ce4e727a16583b5d2ae66ed731296e7fc413b33e8832L415)

### DIFF
--- a/packages/react/cache.ts
+++ b/packages/react/cache.ts
@@ -14,4 +14,4 @@ const cache = {
   }
 };
 
-export = cache;
+export default cache;
--- a/packages/react/cache.ts
+++ b/packages/react/cache.ts
@@ -1,12 +1,6 @@
 const _cache: { [key: string]: string } = {};
 
-interface Cache {
-  set(key: string, value: string): void;
-  get(key: string): string;
-  clear(): void;
-}
-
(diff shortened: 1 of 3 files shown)


packages/react/cache.ts
```diff
@@ -0,0 +1,17 @@
...
+    return _cache[key];
+  },
+  clear(): void {
+    Object.keys(_cache).forEach(key => {
+      delete _cache[key];
+    });
+  }
+};
+
+export = cache;
```
michael-siek, line 17:
```suggestion
export default cache;
```

packages/react/index.ts
```diff
@@ -2,6 +2,7 @@
 import axeCore = require('axe-core');
 import rIC = require('requestidlecallback');
 import after = require('./after');
+import cache = require('./cache');
```
michael-siek, line 5:
```suggestion
import cache from './cache';
```

michael-siek (dismissed):

michael-siek (approved):

PR #499, opened 2022-04-07
test: playwright, puppteer, WDIO, and, WDJS tests

Zidious:
Mostly, the pathing of the test fixtures were incorrect, so it was not testing the correct page.

### DIFF
--- a/packages/playwright/tests/axe-playwright.spec.ts
+++ b/packages/playwright/tests/axe-playwright.spec.ts
@@ -56,7 +56,7 @@ describe('@axe-core/playwright', () => {
 
   describe('analyze', () => {
     it('returns results using a different version of axe-core', async () => {
-      await page.goto(`${addr}/index.html`);
+      await page.goto(`${addr}/external/index.html`);
       const results = await new AxeBuilder({
         page,
         axeSource: axeLegacySource
@@ -70,7 +70,7 @@ describe('@axe-core/playwright', () => {
     });
 
     it('returns results', async () => {
-      await page.goto(`${addr}/index.html`);
+      await page.goto(`${addr}/external/index.html`);
       const results = await new AxeBuilder({ page }).analyze();
       assert.isNotNull(results);
       assert.isArray(results.violations);
@@ -137,7 +137,7 @@ describe('@axe-core/playwright', () => {
       let error: Error | null = null;
 
       const brokenSource = axeSource + `;window.axe.utils = {}`;
-      await page.goto(`${addr}/external/index.html`);
+      await page.goto(`${addr}/external/external/index.html`);
       try {
         await new AxeBuilder({ page, axeSource: brokenSource })
           .withRules('label')
@@ -150,7 +150,7 @@ describe('@axe-core/playwright', () => {
     });
 
     it('returns the same results from runPartial as from legacy mode', async () => {
-      await page.goto(`${addr}/nested-iframes.html`);
+      await page.goto(`${addr}/external/nested-iframes.html`);
       const legacyResults = await new AxeBuilder({
         page,
         axeSource: axeSource + axeForceLegacy
@@ -160,7 +160,17 @@ describe('@axe-core/playwright', () => {
       const normalResults = await new AxeBuilder({ page, axeSource }).analyze();
       normalResults.timestamp = legacyResults.timestamp;
       normalResults.testEngine.name = legacyResults.testEngine.name;
-      assert.deepEqual(normalResults, legacyResults);
+
+      /**
+       * we need to stringify and parse the results as
+       * deep equal thinks "messageKey": [undefined] and
+       * "shadowColor": [undefined] exist, but they do not appear
+       * in the results
+       */
+      assert.deepEqual(
+        JSON.parse(JSON.stringify(normalResults)),
+        JSON.parse(JSON.stringify(legacyResults))
+      );
     });
   });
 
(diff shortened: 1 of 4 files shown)


straker (dismissed):
Since this problem was caused by us not realizing that the pages we navigated to were returning 404s, I think we should take the extra time here and assert that the page is what we expect before running tests. Something like assert the title of the page or the status of the page should be sufficient.

```js
const title = await page.title();
asset.equal(title, 'Nested Frames')

// or

const response = await page.goto(url);
asset.equal(response.status(), 200) 
```


_Edit:_ I think asserting the status is probably the better solution

Zidious:
Unfortunately, [WDJS](https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/ie_exports_Driver.html#get)' `get` returns `undefined`, doesn't appear to be a way to get the status code of a page (technically the page loaded, just with a 404). 

Similarly, WDIO `url` returns a `string[]` which incorrect url path or not returns `null`, doesn't appear to be a way to check the status here either.

packages/playwright/tests/axe-playwright.spec.ts
```diff
@@ -80,7 +84,7 @@ describe('@axe-core/playwright', () => {
     });
 
     it('returns correct results metadata', async () => {
-      await page.goto(`${addr}/external/index.html`);
+      const res = await page.goto(`${addr}/external/index.html`);
```
straker, line 87: `res` isn't used in the test. surprised that didn't throw a lint error for unused variable.

packages/playwright/tests/axe-playwright.spec.ts
```diff
@@ -96,17 +100,19 @@ describe('@axe-core/playwright', () => {
...
         err = e;
       }
+
+      assert.equal(res?.status(), 200);
       assert.isUndefined(err);
     });
 
     it('reports frame-tested', async () => {
-      await page.goto(`${addr}/external/crash-parent.html`);
+      const res = await page.goto(`${addr}/external/crash-parent.html`);
```
straker, line 115: Not used here either

packages/playwright/tests/axe-playwright.spec.ts
```diff
@@ -384,19 +424,20 @@ describe('@axe-core/playwright', () => {
...
       const flattenTarget = flatPassesTargets(results);
 
+      assert.equal(res?.status(), 200);
       assert.strictEqual(flattenTarget[0], '.include');
       assert.notInclude(flattenTarget, '.exclude');
     });
 
     it('with only include', async () => {
-      await page.goto(`${addr}/context.html`);
+      const res = await page.goto(`${addr}/context.html`);
```
straker, line 440: Oy! mate! use res ("here too" in British English)

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -307,8 +332,11 @@ describe('AxePuppeteer', function () {
       });
 
       it('gives a helpful error', done => {
-        const gotoP = page.goto(`${addr2}/index.html`);
-        gotoP.catch(() => ({})); // suppress Node error
+        page.goto(`${addr2}/external/index.html`).then(res => {
+          assert.equal(res.status(), 200);
+        });
+
+        // gotoP.catch(() => ({})); // suppress Node error
```
straker, line 339: Should this be uncommented now?

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -412,7 +442,7 @@ describe('AxePuppeteer', function () {
           }
         `;
 
-        await page.goto(`${addr}/context.html`);
+        const res = await page.goto(`${addr}/context.html`);
```
straker, line 445: aquí también ("here too" in Spanish)

straker (changes requested):
For wdjs and wdio you can assert that the `title` isn't `"Error"`.

```js
await client.url(`${addr}/doesnt-exist-frameset.html`);
const title = await client.getTitle();  // { title: 'Error' }
assert.notEqual(title, 'Error');
```

packages/webdriverjs/tests/axe-webdriverjs.spec.ts
```diff
@@ -91,24 +100,30 @@ describe('@axe-core/webdriverjs', () => {
...
       assert.isUndefined(err);
     });
 
     it('throws if axe errors out on the top window', done => {
       driver
         .get(`${addr}/external/crash.html`)
         .then(() => {
+          driver.getTitle().then(res => {
+            assert.notEqual(res, 'Error');
+          });
```
straker, line 126: This is a race condition since the  code isn't waiting for the response  before going to the analyze.

packages/webdriverjs/tests/axe-webdriverjs.spec.ts
```diff
@@ -121,6 +136,10 @@ describe('@axe-core/webdriverjs', () => {
       driver
         .get(`${addr}/external/crash-me.html`)
         .then(() => {
+          driver.getTitle().then(res => {
```
straker, line 139: Here too

packages/webdriverjs/tests/axe-webdriverjs.spec.ts
```diff
@@ -134,6 +153,10 @@ describe('@axe-core/webdriverjs', () => {
       driver
         .get(`${addr}/external/index.html`)
         .then(() => {
+          driver.getTitle().then(res => {
```
straker, line 156: Here too

straker (dismissed):

straker (approved):

PR #509, opened 2022-05-16
chore: update webdriverio peerDependency

straker:
We believe the issue that caused us to pin to 7.14.1 was this one https://github.com/webdriverio/webdriverio/issues/7994. Testing versions of webdriverio 7.14 - 7.19 no longer causes us any issues. The issue seems to have been resolved and we no longer need to pin to a specific version.


michael-siek:
This PR should be fixed with this https://github.com/dequelabs/axe-core-npm/pull/514

michael-siek (approved):

Issue #519, opened 2022-06-02
Missing cache file in dist


Kosai106:
```
ERROR in ./node_modules/@axe-core/react/dist/index.js
Module not found: Error: Can't resolve './cache' in '/Users/**/**/**/node_modules/@axe-core/react/dist'
 @ ./node_modules/@axe-core/react/dist/index.js 20:30-48
 @ ./src/index.tsx
```

**Product:** `react`

**Expectation:**
I expect the file to be present so my application will run.

**Actual:**
The file is missing from the `node_modules` meaning my app is not able to compile and start.

**Motivation:**
`4.4.0` works but doesn't support React 17
`4.4.2` doesn't work because of this issue https://github.com/dequelabs/axe-core-npm/issues/479, but solves React 17 support...

Hence why we use `@axe-core/react@next` 🙃

```bash
axe-core version: 4.4.1 # This is defined in your own package
@axe-core/react: 4.4.3-alpha.288

- Node version: 16.13.0
- Platform: MacOS Monterey
```

michael-siek:
Hey @Kosai106 Thank you very much for the contribution! this was released in 4.4.3.

padmavemulapati:
cache files observed and not finding error with ../react/dist/index.js. 
Tests are working as expected
<img width="288" alt="image" src="https://user-images.githubusercontent.com/47409579/184813522-aba1ed4f-d04d-4568-91fe-3f1411c3815d.png">

PR #520, opened 2022-06-02
feat: include missing cache file

Kosai106:
https://github.com/dequelabs/axe-core-npm/issues/519

I've verified that this works manually by placing the file in the `node_modules` myself, after which everything runs as intended.


michael-siek:
@Kosai106 Thank you very much for the contribution!

Issue #534, opened 2022-06-24
Support puppeteer version 15


michael-siek:
Puppeteer version 15 is out and we should update to support the update

padmavemulapati:
Validated with the below list of dependencies:
```
  "dependencies": {
    "@axe-core/puppeteer": "^4.4.4-alpha.298",
    "axe-core": "^4.4.2",
    "puppeteer": "^15.0.1"
  } 
```
and working fine and seeing the expected results

PR #535, opened 2022-06-24
fix(puppeteer): support puppeteer v15

michael-siek:
See: #534

### DIFF
--- a/packages/puppeteer/src/legacy.ts
+++ b/packages/puppeteer/src/legacy.ts
@@ -1,5 +1,5 @@
 import * as fs from 'fs';
-import { Frame, ElementHandle } from 'puppeteer';
+import { Frame } from 'puppeteer';
 
 interface IInjectAxeArgs {
   source?: string | Function;
@@ -33,7 +33,7 @@ export async function injectJS(
     console.error(`Failed to inject axe-core into frame (${frame.url()})`);
   };
 
-  let injectP: Promise<ElementHandle<Element> | void>;
+  let injectP: Promise<void>;
   if (!source) {
     injectP = injectJSModule(frame);
   } else {
--- a/packages/puppeteer/src/legacy.ts
+++ b/packages/puppeteer/src/legacy.ts
@@ -61,7 +61,7 @@ function injectJSSource(
   source: string | Function,
   args: any[] = []
 ): Promise<void> {
-  return frame.evaluate(source as any, ...args);
+  return frame.evaluate(source as any, ...args) as Promise<void>;
 }
 
 export function iframeSelector(disabledFrameSelectors: string[]): string {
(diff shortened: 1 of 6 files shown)


packages/puppeteer/src/legacy.ts
```diff
@@ -61,7 +61,7 @@ function injectJSSource(
   source: string | Function,
   args: any[] = []
 ): Promise<void> {
-  return frame.evaluate(source as any, ...args);
+  return frame.evaluate(source as any, ...args) as Promise<void>;
```
michael-siek, line 64: Needed to typecast since this was type `Promise<unknown>`

packages/puppeteer/tsconfig.json
```diff
@@ -18,7 +18,10 @@
     "noImplicitThis": true,
     "noImplicitReturns": true,
     "moduleResolution": "node",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "paths": {
```
stephenmathieson, line 22: Why was this necessary?

michael-siek: Without this it uses the types provided by puppeteer which doesn't export `JSONArray, JSONObject, ElementHandle` but the types file does

michael-siek: We use types that are not exported by puppeteer and use `@types/puppeteer`

stephenmathieson: Where do the types come from? I still don't understand. Why does adding `paths` change anything? Isn't this default behavior?

straker: Don't we usually fix this using the `typesRoot` property?

```
"typeRoots": ["node_modules/@types"]
```

stephenmathieson: From https://github.com/puppeteer/puppeteer/pull/8547:


> [BREAKING CHANGE] We've removed the ability to declare return types in type arguments for the following reasons:
> - Setting them will indubitably break auto-deduction.
> - You can just use `as ...` in TypeScript to coerce the correct type (given it makes sense).

Why do we have to hack TS rather than casting (`as ...`)?

stephenmathieson: That `typeRoots` value is also the default. Specifying it does nothing.

stephenmathieson: I've solved this without any hacks: https://github.com/dequelabs/axe-core-npm/pull/535/commits/b30f8a6a6de52bb6eb5c8aa6b23696eaccf4d529

stephenmathieson (commented):
I should not review this, as I've made changes here.

IMO this should show up in release notes, either as a bug fix or a feature.

We should _always_ look for alternatives to hacking `tsconfig`. By using the _actual_ Puppeteer types, we can avoid all of the hacks/casts and just pass plain objects around.

straker:
Did some quick testing using puppeteer v12-15. All ran without issues (including with typescript).

Issue #553, opened 2022-08-24
CLI: add `--show-errors` as a default behavior 


michael-siek:
Currently, we do not provide a stack trace of the errors and only provide a non-helpful error message. We should instead default it to show the stack trace when the CLI errors.

SantoshDeque:
Verified the ticket with axe version 4.4.5-alpha.312+5d45b8c and below are the observations
1)By default it is giving stack trace or errors as expected 
<img width="980" alt="Screenshot 2022-09-14 at 8 24 12 PM" src="https://user-images.githubusercontent.com/45222788/190191067-033bcd35-7cce-4380-8d42-3165983caf94.png">

2)But, If we pass any value to --show-errors, then it is considering it as false and not showing stack trace of errors
for ex: `axe test.html --show-errors true` or `axe test.html --show-errors someString`
<img width="646" alt="Screenshot 2022-09-14 at 8 30 02 PM" src="https://user-images.githubusercontent.com/45222788/190191097-88b0320b-03e6-4edf-8e02-292a7debc2f8.png">

padmavemulapati:
We are closing this ticket, as the issue is resolved here but we are seeing the behaviour of fix is defaulting for either the value i.e., true/false and tracking the issue [here](https://github.com/dequelabs/axe-core-npm/issues/574)

PR #555, opened 2022-08-25
fix(cli): display stack trace on errors by default

michael-siek:
See: #553


straker (dismissed):
The pr title should be a little more explicit for the changelog

straker (approved):

Issue #382, opened 2021-09-20
Make error-handling url configurable 


straker:
In order to get axe-core-npm 4.3.3 released on time, we opted for a less-than-ideal strategy to hard-code the error-handling url in each package. If instead we allowed the url to be configurable, it would make for a cleaner integration.

```js
this.errorUrl = 'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverjs/error-handling.md';

analyze() {
  try {
  
  } catch(err) {
    throw new Error(`${err.message}. Please see ${this.errorUrl}`)
  }
}
``

somaalapati:
@michael-siek @Zidious Can you please provide some QA notes to test this ticket?

michael-siek:
@somaalapati This is a dev task I don't think there is anything to QA

PR #557, opened 2022-08-26
chore: make errorUrl configurable

michael-siek:
See: #382

### DIFF
--- a/packages/playwright/tests/axe-playwright.spec.ts
+++ b/packages/playwright/tests/axe-playwright.spec.ts
@@ -647,11 +647,6 @@ describe('@axe-core/playwright', () => {
         'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md'
       );
     });
-    it('return modified errorUrl', () => {
-      const builder = new AxeBuilder({ page }) as any;
-      builder.errorUrl = 'https://deque.biz';
-      assert.equal(builder.errorUrl, 'https://deque.biz');
-    });
   });
 
   describe('setLegacyMode', () => {
--- a/packages/playwright/tests/axe-playwright.spec.ts
+++ b/packages/playwright/tests/axe-playwright.spec.ts
@@ -584,6 +584,42 @@ describe('@axe-core/playwright', () => {
           (err as Error).message,
           /Please make sure that you have popup blockers disabled./
         );
+        assert.include(
+          (err as Error).message,
+          'Please check out https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md'
+        );
+      }
+    });
+
+    it('throw an error with modified url', async () => {
+      const res = await page.goto(`${addr}/external/index.html`);
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      //@ts-ignore
+      delete page.context().newPage;
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      //@ts-ignore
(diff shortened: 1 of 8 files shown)


straker (changes requested):
Please add test for the changes

straker (dismissed):
These tests don't test that `errorUrl` is used, just that setting it returns itself. A better test would be to have an error occur that triggers the code where `errorUrl` is used, and check to make sure the url is correct. [We should already have one we can copy](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/tests/axe-playwright.spec.ts#L590-L604).

packages/playwright/tests/axe-playwright.spec.ts
```diff
@@ -603,6 +639,21 @@ describe('@axe-core/playwright', () => {
...
+      assert.equal(
+        errorUrl,
+        'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/error-handling.md'
+      );
+    });
+    it('return modified errorUrl', () => {
+      const builder = new AxeBuilder({ page }) as any;
+      builder.errorUrl = 'https://deque.biz';
+      assert.equal(builder.errorUrl, 'https://deque.biz');
+    });
```
straker, line 654: This isn't a useful test (checking if the value you just set is equal to the value you set)

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -120,6 +120,22 @@ describe('AxePuppeteer', function () {
...
+        errorUrl,
+        'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/puppeteer/error-handling.md'
+      );
+    });
+
+    it('returns modified errorUrl', () => {
+      const builder = new AxePuppeteer(page) as any;
+      builder.errorUrl = 'https://deque.biz';
+      assert.equal(builder.errorUrl, 'https://deque.biz');
+    });
```
straker, line 136: Ditto

packages/webdriverio/src/test.ts
```diff
@@ -141,6 +141,22 @@ describe('@axe-core/webdriverio', () => {
...
+              errorUrl,
+              'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverio/error-handling.md'
+            );
+          });
+
+          it('returns modified errorUrl', () => {
+            const builder = new AxeBuilder({ client }) as any;
+            builder.errorUrl = 'https://deque.biz';
+            assert.equal(builder.errorUrl, 'https://deque.biz');
+          });
```
straker, line 157: Tambien (also in Spanish)

packages/webdriverjs/tests/axe-webdriverjs.spec.ts
```diff
@@ -169,6 +169,22 @@ describe('@axe-core/webdriverjs', () => {
...
+        errorUrl,
+        'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverjs/error-handling.md'
+      );
+    });
+
+    it('returns modified errorUrl', () => {
+      const builder = new AxeBuilder(driver) as any;
+      builder.errorUrl = 'https://deque.biz';
+      assert.equal(builder.errorUrl, 'https://deque.biz');
+    });
```
straker, line 185: もう 一回 (one more time in Japanese)

straker (changes requested):

straker (approved):

Issue #536, opened 2022-06-30
Depend on playwright-core rather on playwright


mxschmitt:
Currently axe playwright depends on playwright. This means that if you install axe playwright, that all browsers get automatically downloaded, since we download the browsers in the postinstall step.

Since around half a year or so, playwright depends on playwright-core and @playwright/test depends on playwright-core.

I'd recommend to also depend on playwright-core instead, which will give you the correct types, but keep the flexibility up to the user which browser he wants to install. Also @playwright/test does not install all the browsers anymore, you need to explicitly call npx playwright install to install them or just one of them.

See here: https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/package.json

michael-siek:
Hey @mxschmitt

Thank you for submitting the recommendation we will discuss this and update the package as needed.

padmavemulapati:
Validated this using `@axe-core/playwright-4.4.5-alpha.313` 
Here while validating, instead of installing `playwright`, installed `playwright-core_1.25.2` (so `playwright-` not exists in the dependencies list)
Run the `js` script
```
const AxeBuilder = require('@axe-core/playwright').default;
const playwright = require('playwright-core');

(async () => {
  const browser = await playwright.chromium.launch();
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto('https://dequeuniversity.com/demo/mars/');

  try {
    const results = await new AxeBuilder({ page }).analyze();
    console.log(results);
  } catch (e) {
    // do something with the error
  }
  await browser.close();
})();
```
and it is giving expected results holding `axe-core_4.4.3`
<img width="1045" alt="image" src="https://user-images.githubusercontent.com/47409579/190981027-0347fe35-a7b4-4663-b76f-ee60b4766073.png">

But here observed one thing:
Even If I pass `const playwright = require('playwright'); ` instead of `const playwright = require('playwright-core');` seeing the expected results.

PR #563, opened 2022-09-13
fix: change `playwright` to devdep and use `playwright-core` types

michael-siek:
See: #536, #536

### DIFF
--- a/packages/playwright/package.json
+++ b/packages/playwright/package.json
@@ -39,8 +39,7 @@
     "prepare": "npm run build"
   },
   "dependencies": {
-    "axe-core": "^4.4.2",
-    "playwright": ">= 1.0.0"
+    "axe-core": "^4.4.2"
   },
   "devDependencies": {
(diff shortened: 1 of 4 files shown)


packages/playwright/package.json
```diff
@@ -53,6 +52,8 @@
     "express": "^4.18.1",
     "mocha": "^10.0.0",
     "nyc": "^15.1.0",
+    "playwright": "^1.0.0",
+    "playwright-core": "^1.25.2",
```
Zidious, line 56: `playwright-core` probably needs to be a dependency as we are utilising functions from Playwright rather than just exposing their types

Zidious (changes requested):
One thing inline 


nit - we may also want to use the types from `playwright-core` in [axe-playwright.spec.ts](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/playwright/tests/axe-playwright.spec.ts) too

Zidious (approved):

PR #566, opened 2022-09-20
fix: only set allowedOrigin when needed

michael-siek:


packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -953,4 +953,44 @@ describe('AxePuppeteer', function () {
...
+    it('should not set when running runPartial and legacy mode', async () => {
+      await page.goto(`${addr}/index.html`);
+      await new AxePuppeteer(page).setLegacyMode(true).analyze();
+      const allowedOrigins = await getAllowedOrigins();
+      assert.deepEqual(allowedOrigins, [addr]);
+    });
+
+    it('should not set when running legacy source and legacy mode', async () => {
+      await page.goto(`${addr}/index.html`);
+      await new AxePuppeteer(page, axeSource + axeForceLegacy)
```
straker, line 981: Odd, doesn't look like this file ever makes the legacy source available outside of the `without run partial` describe.

packages/puppeteer/test/axePuppeteer.test.ts
```diff
@@ -953,4 +953,44 @@ describe('AxePuppeteer', function () {
...
+        .setLegacyMode(true)
+        .analyze();
+      const allowedOrigins = await getAllowedOrigins();
+      assert.lengthOf(allowedOrigins, 1);
```
straker, line 985:
```suggestion
      assert.deepEqual(allowedOrigins, [addr]);
```

packages/webdriverio/src/index.ts
```diff
@@ -224,6 +231,10 @@ export default class AxeBuilder {
     if (!runPartialSupported || this.legacyMode) {
       return await this.runLegacy(context);
     }
+
+    if (!this.legacyMode && !runPartialSupported) {
+      await configureAxe(this.client);
+    }
```
straker, line 237: At this point we know run partial is supported so don't need to configure

```suggestion
```

packages/webdriverio/src/index.ts
```diff
@@ -295,7 +306,13 @@ export default class AxeBuilder {
...
         assert(frame, `Expect frame of "${frameSelector}" to be defined`);
         await this.client.switchToFrame(frame);
-        await axeSourceInject(this.client, this.script);
+        const runPartialSupported = await axeSourceInject(
+          this.client,
+          this.script
+        );
+        if (!this.legacyMode && !runPartialSupported) {
+          await configureAxe(this.client);
+        }
```
straker, line 315: We're in run partial so it's supported.

```suggestion
```

packages/webdriverio/src/utils.ts
```diff
@@ -183,6 +183,14 @@ export const axeFinishRun = (
   );
 };
 
+export const configureAxe = (client: Browser<'async'>): Promise<void> => {
```
straker, line 186: This name is a bit of a misnomer. Would be better as `configureAllowedOrigins`

```suggestion
export const configureAllowedOrigins = (client: Browser<'async'>): Promise<void> => {
```

packages/webdriverjs/tests/axe-webdriverjs.spec.ts
```diff
@@ -861,4 +861,42 @@ describe('@axe-core/webdriverjs', () => {
...
+      assert.deepEqual(allowedOrigins, [addr]);
+    });
+
+    it('should not set when running legacy source and legacy mode', async () => {
+      await driver.get(`${addr}/index.html`);
+      await new AxeBuilder(driver, axeSource + axeForceLegacy)
+        .setLegacyMode(true)
+        .analyze();
+      const allowedOrigins = await getAllowedOrigins();
+      assert.lengthOf(allowedOrigins, 1);
```
straker, line 891:
```suggestion
      assert.deepEqual(allowedOrigins, [addr]);
```

straker (changes requested):

straker (approved):

PR #606, opened 2022-11-14
ci: create semantic pr title GHA

michael-siek:


stephenmathieson (approved):
We should probably move this into a standalone action repository someday. It's been copy/pasted into dozens of places now.

PR #607, opened 2022-11-15
fix: update to use axe-core 4.5.2

michael-siek:


packages/react/examples/shadow-dom/package-lock.json
```diff
@@ -9,7 +9,6 @@
       "version": "4.5.1",
       "license": "MIT",
       "dependencies": {
-        "@axe-core/react": "^4.5.1",
```
Zidious, line 12: axe-core appears to be removed rather than updated to 4.5.2

Zidious (changes requested):
Minor comment about the React shadow dom example package otherwise LGTM

packages/react/examples/shadow-dom/package-lock.json
michael-siek: axe-core isn't a dependency of this package

Zidious: Oh my mistake

Zidious (approved):

PR #632, opened 2022-12-14
fix: use `SerialContextObject` type instead of `ContextObject`

michael-siek:


packages/webdriverio/src/utils.ts
```diff
@@ -183,7 +184,9 @@ export const axeFinishRun = (
   );
 };
 
-export const configureAllowedOrigins = (client: Browser<'async'>): Promise<void> => {
+export const configureAllowedOrigins = (
+  client: Browser<'async'>
+): Promise<void> => {
```
michael-siek, line 189: feel free to ignore this change this was just prettier doing things

straker (commented):
Do we need to update the types for context in https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverjs/src/index.ts?

straker (approved):

PR #652, opened 2023-01-24
chore: update puppeteer and axe-core versions

colinrotherham:
Hello 👋

Appreciate this feels like an unsolicited `npm update` (sorry)

Would you mind if we bumped the `puppeteer` and `axe-core` versions?

For context, I've been working on automated test tooling work over on:

* https://github.com/alphagov/govuk-frontend/issues/1971
* https://github.com/alphagov/govuk-frontend/issues/3041

Which led to quite a major Puppeteer version bump for our [GOV.UK Design System](https://github.com/alphagov/govuk-design-system) tests:

* https://github.com/alphagov/govuk-design-system/pull/2532

Colour contrast checks in `axe-core` are something we're keen to test more across all our [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) components so this update would align with our use of `puppeteer@19` on that repo ✅ 

Possible concerns to note down, I'm not sure if the recent [Chromium v110 bump](https://github.com/puppeteer/puppeteer/pull/9500) will require `chromedriver@110` to hit stable?

Thanks 😊

PR #655, opened 2023-01-31
refactor: Remove lockfiles

stephenmathieson:
Keeping `package-lock.json` files here causing [contributor problems](https://github.com/dequelabs/axe-core-npm/pull/652#issuecomment-1408801982) and prevents CI from testing with the latest versions of our dependencies. They provide no value, so they're being removed.


colinrotherham:
@stephenmathieson Thanks for looking at this

Lerna now [recommends using workspaces](https://github.com/lerna/lerna/blob/116c62c94209950436880fd077ffc5bcfd8fe7b3/website/docs/getting-started.md?plain=1#L94) if that's easier?

>#### [useWorkspaces & packages​](https://lerna.js.org/docs/api-reference/configuration#useworkspaces--packages)
>
>Since Lerna was created, all major package managers (npm, yarn, and pnpm) have added the ability to cross-link packages in the same repo and dedupe node modules. If you'd like Lerna to delegate this process to the package manager you use, set `useWorkspaces: true` in `lerna.json`.

Lots of benefits such as:

1. Project level `package-lock.json` only
2. Project level `npm audit` works again
3. Much faster install (including `npm ci`)

With `lerna bootstrap` being deprecated you'll find `lerna link` isn't needed either (npm does it instead)

### Lerna config
Add `{ "useWorkspaces": true }` to let npm install/link/hoist dependencies

### npm config
Add `{ "workspaces": [] }` so npm starts managing child dependencies

```json
"packages": [
  "packages/*",
  "packages/react/examples/*",
  "packages/webdriverjs/tests/example",
  "axe_core_test"
],
```

stephenmathieson:
@colinrotherham Thanks for the information. However, the idea here is to remove the lockfiles entirely. They only make work work for us.

colinrotherham:
@stephenmathieson Ah so even a single project lockfile is too much?

Don't worry

straker:
Closed via switching to npm workspaces #687. If we still feel the package-lock file presents a problem, we should open an issue to discuss it.

PR #667, opened 2023-02-09
fix: work with extremely large results

### LINKED ISSUE #666: Execute script for all drivers does not work when passed an extremely large results
The `execute` command for each driver has a maximum allowed character / size length before the driver errors. We need to change how we are injecting the partial for `finishRun` in order to support extremely large results.

straker:
Closes #666


packages/playwright/src/browser.ts
```diff
@@ -36,8 +37,12 @@ export const axeRunPartial = ({
...
-  partialResults,
   options
 }: FinishRunParams): Promise<AxeResults> => {
-  return window.axe.finishRun(partialResults, options);
+  return window.axe.finishRun(JSON.parse(window.partialResults), options);
 };
+
+export function chunkResultString(chunk: string) {
+  window.partialResults ??= '';
+  window.partialResults += chunk;
```
stephenmathieson, line 47: What happens if the site under test has this global defined? We should at least prefix this with "axe".

packages/playwright/src/browser.ts
```diff
@@ -36,8 +37,12 @@ export const axeRunPartial = ({
 };
 
 export const axeFinishRun = ({
-  partialResults,
   options
 }: FinishRunParams): Promise<AxeResults> => {
-  return window.axe.finishRun(partialResults, options);
+  return window.axe.finishRun(JSON.parse(window.partialResults), options);
```
stephenmathieson, line 42: Why can't `partialResults` be passed in here?

stephenmathieson (changes requested):
I'm _very_ worried about the new global.

packages/playwright/src/browser.ts
straker: That was the problem. It was too big to be put here on it's own

packages/playwright/src/browser.ts
stephenmathieson: Intersting, OK 👍

stephenmathieson (approved):

Issue #660, opened 2023-02-08
Unable to import defaults when using .mjs


WilcoFiers:
For some reason I'm not getting the default returned when importing `@axe-core/playwright` with MJS. The only way to get it working is to do this:

```js
// test.mjs
import axe from '@axe-core/playwright'
const AxeBuilder = axe.default;
```

I didn't check, but I imagine other packages will have this same issue.

straker:
To solve this, we'll need to build a mjs version along side our current version and have the package.json file point to both of them (current version as the "main" property, mjs version as the "module" property).

So the ticket to support esm in our repos should be these steps:

* add named export for whatever the default export is (with test)
* compile an mjs version along side our current build step
* update package.json to use module property to point to the mjs file

IGx89:
With axe-core 4.7.0 just releasing, a feature release, does that mean this is close to being resolved? This issue has been causing us a lot of annoyance as well (forced to turn off type checking on our tests) so eagerly awaiting a fix :)

straker:
We're working to add this to the 4.7 release of the apis, but I can't say for certain if it will make it in.

padmavemulapati:
Validated with the latest axe-core-npm version, 4.7.1-alpha.383 
We were blocked with [an issue](https://github.com/dequelabs/axe-core-npm/issues/717), once the issue resolved will validate this ticket

padmavemulapati:
Valdiated with the latest @axe-core-npm packages (4.6.2-alpha.380)

So we have now all the `npm` packages,  supports esm in our repo and having the below:

1. added named export for  the default export 
2.current build getting  compiled mjs version along side 
3. package.json  using module property to point to the mjs file

for ex: axe-core/playwright is looking like below:
![Image](https://user-images.githubusercontent.com/47409579/236398559-afe585fa-f8c9-4daa-bdfe-929fc0c4a452.png)

michael-siek:
Closing please reopen if relevant

PR #668, opened 2023-02-15
chore(playwright): add named export 

AdnoC:
https://github.com/dequelabs/axe-core-npm/issues/660

Added a named export while leaving the default export intact. This allows code that was previously using playwright to continue to do so, but allow new code to use named exports regardless of whether it is using `require` or `import` style imports.

The following .mjs file properly imports playwright:

```javascript
// test.mjs
import { AxeBuilder } from '@axe-core/playwright';
console.log(AxeBuilder);
```

### DIFF
--- a/packages/playwright/src/index.ts
+++ b/packages/playwright/src/index.ts
@@ -22,7 +22,7 @@ import {
 } from './browser';
 import AxePartialRunner from './AxePartialRunner';
 
-export default class AxeBuilder {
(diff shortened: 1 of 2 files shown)


straker (changes requested):
I think we should go with the solution in https://github.com/dequelabs/axe-core-npm/pull/671 instead of creating a named export.

Issue #670, opened 2023-02-17
Distribute `@axe-core/playwright` for both CJS and ESM


akornmeier:
**Product:** `axe-core/playwright`

**Expectation:** Package distributes both CJS and ES modules

**Actual:** Package only distributed CJS module which causes a few issue when imported to a project using ES modules by default (like ViteJS referenced in issue #601). 

**Motivation:** ESM has been an out-of-box feature in Node since v14. Supporting projects that use that technology helps keeping the web accessible by using these great packages. 

<pre><code>
axe-core version: v4.6.3
@axe-core/playwright: v4.6.0

- Node version: v16.14.2
- Platform:  Mac
</code></pre>

Zidious:
Hey Tony ☕ ,

This should be resolved in the `@latest` version. 

Please let me know if you run into any issues!

michael-siek:
Closing please reopen if relevant. 

Also Hi @akornmeier! 🥂

padmavemulapati:
Verified during 4.7.2 rlease testing and now @axe-core/playwright is supporting both CJS and ESM

Issue #601, opened 2022-11-07
@axe-core/playwright throws an `TypeError: AxeBuilder is not a constructor`  error


Volubyl:
Product:  @axe-core/playwright
Expectation: 

Should do something other than throw an `TypeError: AxeBuilder is not a constructor`  error

Actual: 

I followed the [example of accessibility test provided by Playwright](https://playwright.dev/docs/accessibility-testing#example-accessibility-tests) but I ended up with this error 

```
`TypeError: AxeBuilder is not a constructor`
```
The test looks like :

```
import { test } from '@playwright/test';
import AxeBuilder from '@axe-core/playwright'; // 1

test.describe('homepage', () => { // 2
  test('should not have any automatically detectable accessibility issues', async ({ page }) => {
    await page.goto('/'); // 3

    try {
        const results = await new AxeBuilder({ page }).analyze();
        console.log(results);
      } catch (e) {
         console.log(e) // TypeError: AxeBuilder is not a constructor
      }
  });
});

```

Motivation: 

Because at the moment I can not use axe core with playwright

 - "@axe-core/playwright": "^4.5.0",
 - "@playwright/test": "^1.27.1",

- Node version: 16.17.1
- Platform:  Linux

Volubyl:
I tried to tinker a bit this morning, and at the moment my best guess would be a typescript build "issue".

Maybe something about esm module ?

For the context, I'm working with [SvelteKit](https://kit.svelte.dev/) and I had several issues in the past with ESM modules.

At the moment, I'm only at the beginning of my investigation, so it's only a guess ...

Zidious:
Hey @Volubyl, 

Apologies for the delay - we will investigate this issue and get back to you as soon as possible

Volubyl:
No worries, I was so glad to learn axe core had a playwright integration. I'll try to go further in this investigation.

I cloned the repo to get a local version of the package and was able to reproduce using `npm link` so I'm ready to contribute ;-)

Volubyl:
Ok ... I think you can close this issue. It seems that the issue is not directly related to @axe-core/playwright, but more generally to an ESM/CJS compatibility problem.

Long story short : [Vite](https://vitejs.dev/) (the bundle) "encourage" the use of ESM modules instead of CJS and has sometimes issues to work properly when there is a mix of ESM and CJS module in a single APP.

[...]

PR #671, opened 2023-02-17
feat(playwright): distributes both ESM and CJS modules

akornmeier:
Fixes: [#670, #601]

- add: tsup package to handle building/exporting both ESM and CJS modules
- update: README to include ESM and CJS import/require techniques for users
- update: mocha to work w/ ES modules

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -206,8 +206,6 @@ jobs:
       NODE_ENV: development
     working_directory: ~/axe-core-npm
     steps:
-      # We need to pin node@14 to run WDIO due to WDIO sync, only works on node@14
-      - run: npm install -g node@14 --force
       - checkout
       - restore_dependency_cache
       - run: npm install
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,6 +171,21 @@ jobs:
       - run: npm --prefix=packages/webdriverjs run build
       - run: npm run test --prefix=packages/webdriverjs/tests/example
 
+  playwright:
+    docker:
+      - image: cimg/node:16.16.0-browsers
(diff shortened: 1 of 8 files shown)


stephenmathieson:
Sup Tony 👋

akornmeier:
> Sup Tony 👋

@stephenmathieson - Hey Stephen. 

image

straker (approved):
Thanks for the fixes!

straker:
Approved for security

straker:
Since this is a feat and we have a planned patch release soon, we're going to hold off on merging this for now. Will reevaluate once we get the patch out.

akornmeier:
> Since this is a feat and we have a planned patch release soon, we're going to hold off on merging this for now. Will reevaluate once we get the patch out.

I understand, however I could change change the commit to a fix since this is technically fixing an existing issue and not adding extra functionality. Let me know what you think the best path forward would be? 

What would the ETA for the changes to land if they don't make it into the patch release?

straker:
We want to keep this a `feat` since we are adding official support for modules where we didn't have it before. As for a timeline, axe-core is currently working on a 4.7 release, but I'm not sure on the deadline of it yet, but it will be some time.

straker:
After talking to @AdnoC and @michael-siek, and trying to [update all of our packages to do this technique](678), we've decided to not do this approach as it isn't nearly as nice as we would have hoped and requires converting all the files to CJS. We'll instead work on what a lot of other libraries have done which is moving away from a default export and only supporting named exports.

For now, we'll support `require` using either a default or a named export, but `import` will be a named export only. For the next major release we'll remove the default export for require.

michael-siek:
Closing no longer required

PR #678, opened 2023-03-03
[WIP]feat: distributes both ESM and CJS modules. [ON HOLD]

AdnoC:
https://github.com/dequelabs/axe-core-npm/issues/660

Despite the branch name, this will have changes for all non-playwright packages.

TODO: add tests that this is usable by ts and js import methods. Don't set out of draft without it


straker:
After talking to @AdnoC and @michael-siek, we've decided to not do this approach as it isn't nearly as nice as we would have hoped and requires converting all the files to CJS. We'll instead work on what a lot of other libraries have done which is moving away from a default export and only doing named exports.

PR #687, opened 2023-03-16
refactor: switch to NPM workspaces

not-my-profile:
The `lerna bootstrap` command is deprecated[1]
and resulted in a very weird CI failure (#685).

This commit switches the repository to instead use
NPM workspaces.[2]

[1]: https://github.com/lerna/lerna/discussions/3410
[2]: https://docs.npmjs.com/cli/v7/using-npm/workspaces

Fixes #685.

Issue #685, opened 2023-03-16
ci: dependencies CI job is sometimes failing


not-my-profile:
The `lerna bootstrap` command part of the `dependencies` CI job is currently sometimes failing with the error message:

```
npm ERR! code 1
npm ERR! path /home/circleci/axe-core-npm/packages/webdriverio/node_modules/chromedriver
npm ERR! command failed
npm ERR! command sh -c node install.js
npm ERR! Current existing ChromeDriver binary is unavailable, proceeding with download and extraction.
npm ERR! Downloading from file:  https://chromedriver.storage.googleapis.com/109.0.5414.74/chromedriver_linux64.zip
npm ERR! Saving to file: /tmp/109.0.5414.74/chromedriver/chromedriver_linux64.zip
npm ERR! Received 1040K...
npm ERR! Received 2080K...
npm ERR! Received 3120K...
npm ERR! Received 4160K...
npm ERR! Received 5200K...
npm ERR! Received 6240K...
npm ERR! Received 7129K total.
npm ERR! Extracting zip contents to /tmp/109.0.5414.74/chromedriver.
npm ERR! ChromeDriver installation failed Error: Error extracting archive: Error: unexpected end of file
npm ERR!     at extractDownload (/home/circleci/axe-core-npm/packages/webdriverio/node_modules/chromedriver/install.js:309:11)
npm ERR!     at async install (/home/circleci/axe-core-npm/packages/webdriverio/node_modules/chromedriver/install.js:71:7)
```

This appears to be some unreliable CI fluke since the job failed and succeeded once [for the exact same commit](https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm?branch=pull%2F684):

![image](https://user-images.githubusercontent.com/73739153/225622078-812c5f99-fdb1-4d1e-9788-233b7eb726c3.png)

I am not sure what exactly is causing this CI failure ... it doesn't happen locally for me. Changing the command to `npm run bootstrap -- --ci || (md5sum /tmp/109.0.5414.74/chromedriver/chromedriver_linux64.zip; ls -l /tmp/109.0.5414.74/chromedriver; false)` [revealed that the zip extraction indeed is successful](https://app.circleci.com/pipelines/github/dequelabs/axe-core-npm/1525/workflows/8d80f53e-17b9-4a5d-91da-b7e3c609458c/jobs/12559) ... which makes the occurrence of the "Error extracting archive" all the more puzzling.

[...]

PR #690, opened 2023-03-17
refactor: make tests respect CHROMEDRIVER_PATH

not-my-profile:
This change helps prevent chromedriver version mismatches in the CI
by no longer requiring the chromedriver version in `package-lock.json`
to match the chrome version installed in the CI.

I already previously added similar logic in 3a38cf751636aa9846c77c580e21bb903b7185f6
for webdriverio. This commit improves that logic to no longer hard-code the chromedriver
path in the test and instead provide it via an environment variable (so it can also be used
to run the tests locally against a Chrome version different than the locked chromedriver version).
This commit additionally adapts webdriverjs and webdriverjs-example to also respect
this new CHROMEDRIVER_PATH environment variable.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -155,11 +155,13 @@ jobs:
     steps:
       - checkout
       - restore_dependency_cache
-      # required for browser-driver-manager
-      - run: sudo apt-get update -y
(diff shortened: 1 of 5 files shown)


not-my-profile:
The dependencies test failure is again the CI fluke #685 that will be fixed by the followup PR #687.

straker (approved):
Thanks for the pr! Looks good

Issue #700, opened 2023-03-28
cli: force chrome bin path


sinedied:
Product: cli 

Expectation: 
I'm embedding the CLI in another NPM package pre-configured to work out of the box for our needs. In that context, I would like to install a working Chrome / Chromedriver combo to make sure it works on the target machine, so I would like to be able to force chromedriver and chrome binary paths to be used.

Actual: There's currently an option to force chromedriver path, but not to force chrome binary path.

Motivation: Add an option (or environment variable) so we can force the chrome path to be used.

<pre><code>
@axe-core/cli: 4.6.0

- Node version: XX  v18.14.2
- Platform:  Mac
</code></pre>

straker:
Closed via #704

PR #704, opened 2023-04-03
feat(cli): add --chrome-path option (#700)

sinedied:


Zidious:
Hey @sinedied,

Thanks for the PR! Apologies the delay, we'll get get to reviewing it this week.

packages/cli/src/lib/webdriver.ts
```diff
@@ -21,6 +21,10 @@ const startDriver = async (
       }, options);
     }
 
+    if (config.chromePath) {
+      options.setChromeBinaryPath(config.chromePath);
```
michael-siek, line 25: I don't think we should force a user to type an absolute path here we should do something like 
```js
options.setChromeBinaryPath(path.resolve(config.chromePath));
```

michael-siek (changes requested):

packages/cli/src/lib/webdriver.ts
sinedied: I made the change and rebased the PR, sorry for the delay

packages/cli/src/bin/cli.ts
```diff
@@ -77,6 +77,10 @@ program
     '--chromedriver-path <path>',
     'Absolute path to the desired chromedriver executable'
   )
+  .option(
+    '--chrome-path <path>',
+    'Absolute path to the desired chrome executable'
```
straker, line 82: With the last change we'll need to update the message as well.

```suggestion
    'Path to the desired chrome executable'
```

straker (changes requested):

Zidious:
Hey @sinedied, 

Just checking up on the status of this PR or you need us to take this over? 

Thanks again!

sinedied:
Thanks @Zidious for the reminder!
I forgot about it as I was in vacation during the last notification, oups... I made the requested changes and rebased the branch, should be good now

Zidious:
Hey @sinedied, 

We should have CI up and running on contributor PR's soon and we'll be able to merge in your PR. 

Thanks again!

straker (approved):

straker:
Thanks again for the contribution.

Approved for security.

Issue #705, opened 2023-04-09
@axe-core/webdriverio expects object but get's proxy object


yarinsa:
Product: webdriverio

Expectation: Client should be initialised after running "new AxeBuilder" 

Actual: 
```
AssertionError [ERR_ASSERTION]: An instantiated WebdriverIO client greater than v5 is required
    at new AxeBuilder
```

<pre><code>
@axe-core/webdriverio: 4.6.1
- WDIO: 8.6.9
- Node version: 18.5.0
- Platform:  Mac
</code></pre>

## Description
I am trying to run WDIO accessibility test with axe builder. Using the simple example online. 

```
describe("Accessibility Test", () => {
  it("should get the accessibility results from a page", async () => {
    await browser.url("/iframe.html?id=design-system-button--playground");
 // using AxeBuilder.default() as a workaround until ESM are supported 
 // https://github.com/dequelabs/axe-core-npm/issues/703#issuecomment-1499146493
    const builder = new AxeBuilder.default({ client: browser });
    const result = await builder.analyze();
    console.log("Acessibility Results:", result);
  });
});
```
After running I am getting an error. I debugged the error and the issue is because axe is expecting object and WDIO 8 provides a proxy object. I tried to reverse the browser proxy to object before sending it but it doesn't work .
![Screenshot 2023-04-09 at 15 52 42](https://user-images.githubusercontent.com/44482362/230774097-38856510-d68a-4394-917e-18e919bd41f4.png)
![Screenshot 2023-04-09 at 15 52 32](https://user-images.githubusercontent.com/44482362/230774103-65f816b3-97be-4f4b-9e8c-33a554488b09.png)

singhMo007:
> Thanks for the issue. At this time I believe @axe-core/webdriverio does not support WDIO v8. It's at least not one of our [allowed peer dependencies](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverio/package.json#L66). We probably will look into WDIO v8 support once #695 is completed.

Any idea by when it will be taken care in future?

singhMo007:
Hope it will be taken soon, that will really help who are willing to use axe-core with WDIO V8

padmavemulapati:
Validated with the latest axe-core/webdriverio_4.7.1-alpha.383 , with WDIO version 8.6.7 and node version 18.7.0 - .js tests and APIs are working fine

Issue #695, opened 2023-03-22
remove `@wdio/sync`


michael-siek:
Since `@wdio/sync` no longer works with node 14 and with node 14 being EOL at the end of April we need to also drop support for `@wdio/sync` in our `@axe-devtools/webdriverio` package. We need to also make sure that we run test in CI to support node 18 with the `@axe-devtools/webdriverio` package.

padmavemulapati:
Validated with the latest axe-core/webdriverio_4.7.1-alpha.383 , with WDIO version 8.6.7 - `.js` tests and APIs are working fine . Need to validate with `.mjs` 

<img width="724" alt="image" src="https://user-images.githubusercontent.com/47409579/235587286-0c34df8a-9fc3-41b2-9ba9-2af3f571cc18.png">

PR #708, opened 2023-04-14
feat: support WDIO v8

michael-siek:
Closes: https://github.com/dequelabs/axe-core-npm/issues/695
Closes: https://github.com/dequelabs/axe-core-npm/issues/705

### DIFF
--- a/packages/webdriverio/src/index.ts
+++ b/packages/webdriverio/src/index.ts
@@ -21,12 +21,7 @@ import type {
   SerialSelectorList,
   SerialFrameSelector
 } from 'axe-core';
-import type {
-  Options,
-  CallbackFunction,
-  WdioBrowser,
-  PartialResults
-} from './types';
+import type { Options, CallbackFunction, PartialResults } from './types';
 
 export default class AxeBuilder {
   private client: Browser;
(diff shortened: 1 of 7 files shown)


packages/webdriverio/src/index.ts
```diff
@@ -48,7 +46,7 @@ export default class AxeBuilder {
     // Treat everything as Browser<'async'>:
     // - Anything sync can also run async, since JS can await sync functions
     // - Ignore MultiRemoteBrowser, which is just Browser with extra props
-    this.client = client as Browser<'async'>;
+    this.client = client as Browser;
```
Zidious, line 49: If we are casting `client as Browser` then the type: 

```ts
WdioBrowser = Browser | MultiRemoteBrowser;
```

Could just be `import type { Browser } from 'webdriverio` and we could simply remove `WdioBrowser`. Otherwise we could remove the casting and let TypeScript handle whether the user passes in a type of `Browser` or `MultiRemoveBrowser`

Zidious (commented):
Just had a question inline -

straker (approved):
Michael confirmed this works in TS, node 14, and WDIO v7 from a user importing our code so we shouldn't have any backwards compatibility issues and typescript.

Issue #717, opened 2023-05-01
BUG: ESM support is throwing error for Puppeteer and webdriverio


padmavemulapati:
##Steps to reproduce:
1. Installed latest @axe-core/puppetter and @axe-core/webdriverio (npm install @axe-core/puppeteer@next)
2. Imported into test file as need to get support for `.mjs` files
Testfile (test.mjs) for puppeteer
```
import AxeBuilder from '@axe-core/playwright';
import playwright from'playwright';
import fs from'fs';

(async () => {
  const browser = await playwright.chromium.launch();
  const context = await browser.newContext();
  const page = await context.newPage();
  await page.goto('http://qateam.dequecloud.com/attest/api/test.html');

  try {
    const results = await new AxeBuilder({ page }).analyze()
     fs.writeFile("./Testmjs.json", JSON.stringify(results, null,4),(err)=>
    {if(err){
        console.error(err);
        return;
        };
        console.log("File has been created");
    })
  } catch (e) {
    // do something with the error
  }
  await browser.close();
})();
```

and Testfile for @axe-core/webdriverio
```
import AxeBuilder from '@axe-core/webdriverio';
import { remote } from 'webdriverio';
import fs from 'fs';

(async () => {
  const client = await remote({
    logLevel: 'error',
    capabilities: {
      browserName: 'chrome'
    }
  });

  await client.url('http://qateam.dequecloud.com/attest/api/test.html');

  const builder = new AxeBuilder({ client });
  const results = await builder.analyze();
    fs.writeFile("./Testmjs.json", JSON.stringify(results, null,4),(err)=>
    {if(err){
        console.error(err);
        return;
        };
        console.log("File has been created");
    })
   client.deleteSession();
})();
```
3. run this tests using `node --experimental-modules test.mjs`

Product:  puppeteer  & webdriverio 

Expectation: Should successfully run the test and gives the output in specified `.json` file

Actual: 
**Issue on puppeteer**
```
MacBook-Pro:puppeteer dq_user$ node --experimental-modules test.mjs

  Puppeteer old Headless deprecation warning:
    In the near feature `headless: true` will default to the new Headless mode
    for Chrome instead of the old Headless implementation. For more
    information, please see https://developer.chrome.com/articles/new-headless/.
    Consider opting in early by passing `headless: "new"` to `puppeteer.launch()`
    If you encounter any bugs, please report them to https://github.com/puppeteer/puppeteer/issues/new/choose.

[...]

PR #718, opened 2023-05-01
fix: proper imports in esm context

AdnoC:
Closes: https://github.com/dequelabs/axe-core-npm/issues/717

### DIFF
--- a/packages/puppeteer/src/legacy.ts
+++ b/packages/puppeteer/src/legacy.ts
@@ -59,6 +59,7 @@ async function injectJSModule(frame: Frame): Promise<void> {
     axeCorePath = require.resolve('axe-core');
   } else {
     const { createRequire } = (await import('node:module')) as any;
+    // `getFilename` is needed because esm's `import.meta.url` is illegal syntax in cjs
     const filename = pathToFileURL(getFilename()).toString();
 
     const require = createRequire(filename);
--- a/packages/puppeteer/src/legacy.ts
+++ b/packages/puppeteer/src/legacy.ts
@@ -52,7 +52,17 @@ export async function injectJS(
 }
 
 async function injectJSModule(frame: Frame): Promise<void> {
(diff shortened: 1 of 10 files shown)


packages/puppeteer/src/legacy.ts
```diff
@@ -52,7 +54,18 @@ export async function injectJS(
...
+    axeCorePath = require.resolve('axe-core');
+  } else {
+    const { createRequire } = (await import('node:module')) as any;
+    const filename = pathToFileURL(getFilename()).toString();
+
+    const require = createRequire(filename);
+    axeCorePath = require.resolve('axe-core');
+  }
+
+  const source = fs.readFileSync(axeCorePath, 'utf8');
```
Zidious, line 68: We should comment on why this was necessary to add vs just using `__filename`

packages/webdriverio/src/index.ts
```diff
@@ -23,6 +25,20 @@ import type {
 } from 'axe-core';
 import type { Options, CallbackFunction, PartialResults } from './types';
 
+let axeCorePath = '';
+async function loadAxePath() {
```
Zidious, line 29: nit - we use a function `loadAxePath` here but not puppeteer, IMO we should keep it consistent

Zidious (dismissed):
Minor things inline, also desc of PR should use "Closes: XXX" so it can go into the QA column

packages/webdriverio/src/index.ts
AdnoC: We use a function here because we need to do async stuff at the top-level in order to initialize `axeCorePath`. And for some reason that still isn't allowed. puppeteer does this inside an async function so it isn't needed.

Zidious (dismissed):
Sorry - forgot to add, we should add a test

Zidious (dismissed):

Issue #729, opened 2023-05-17
axe command returns "ENOENT: no such file or directory" error


CarlyGerard:
Product: @axe-core/cli

Expectation: if running <code>axe [url]</code> on the command line (like <code>axe www.github.com</code>), it should start running accessibility checks and return any violations.

Actual: Running <code>axe [url]</code> returns the following error:

<pre><code>
% axe github.com
node:fs:584
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/opt/homebrew/lib/node_modules/node_modules/axe-core/axe.js'
    at Object.openSync (node:fs:584:3)
    at Object.readFileSync (node:fs:452:35)
    at getAxeSource (/opt/homebrew/lib/node_modules/@axe-core/cli/dist/cli.js:106:28)
    at Command.cli (/opt/homebrew/lib/node_modules/@axe-core/cli/dist/cli.js:318:20)
    at Command.listener [as _actionHandler] (/opt/homebrew/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:482:17)
    at /opt/homebrew/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1283:65
    at Command._chainOrCall (/opt/homebrew/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1177:12)
    at Command._parseCommand (/opt/homebrew/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1283:27)
    at Command.parse (/opt/homebrew/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:909:10)
    at Object.<anonymous> (/opt/homebrew/lib/node_modules/@axe-core/cli/dist/cli.js:464:9) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/opt/homebrew/lib/node_modules/node_modules/axe-core/axe.js'
}
</code></pre>

- axe-core version: 4.7.1
- Node version: 19.4.0 
- Platform: macOS (Monterey, 12.6)
- Chrome version: 113.x

Zidious:
Hey @CarlyGerard, 

Looks like you're using Node19 which is the _current_ but not the _current LTS_ version. Using Node18, do you get the same error?

CarlyGerard:
Thanks @Zidious. I downgraded to node18 as suggested, I still seem to be getting the same error:

<pre><code>
% node -v
[...]

Issue #726, opened 2023-05-09
axe-core/cli shouldn't require -a on fresh install


plcj:
Product: cli

Expectation:

Execute `axe https://www.duckduckgo.com` after being installed with `npm install @axe-core/cli -g`

Actual:

```
node:fs:585
  handleErrorFromBinding(ctx);
  ^

Error: ENOENT: no such file or directory, open '/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/node_modules/axe-core/axe.js'
    at Object.openSync (node:fs:585:3)
    at Object.readFileSync (node:fs:453:35)
    at getAxeSource (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/dist/cli.js:106:28)
    at Command.cli (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/dist/cli.js:318:20)
    at Command.listener [as _actionHandler] (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:482:17)
    at /Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1283:65
    at Command._chainOrCall (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1177:12)
    at Command._parseCommand (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:1283:27)
    at Command.parse (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/node_modules/commander/lib/command.js:909:10)
    at Object.<anonymous> (/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/@axe-core/cli/dist/cli.js:464:9) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/Users/me/.nvm/versions/node/v16.15.0/lib/node_modules/node_modules/axe-core/axe.js'
}
```

Motivation:

axe core cli shouldn't require -a after a fresh install

<pre><code>
axe-core version: 4.7.0? if it's different from the @axe-core/cli version, I don't know how to find it
@axe-core/[integration]: 4.7.0

- Node version: v16.15.0  
[...]

PR #730, opened 2023-05-17
fix(packages/cli): revert ESM 

Zidious:
This PR reverts the recent conversion to ESM for CLI as it's not required which should fix the pathing for the CLI binary when looking for the axe-core source.

Closes: https://github.com/dequelabs/axe-core-npm/issues/729
Closes: https://github.com/dequelabs/axe-core-npm/issues/726


packages/cli/src/lib/utils.ts
```diff
@@ -76,30 +76,12 @@ export const getAxeSource = (axePath?: string): string | void => {
...
-      '..',
-      '..',
-      '..',
-      'node_modules',
-      'axe-core',
-      'axe.js'
-    );
-  }
+  const a = require.resolve('axe-core');
+  axePath = path.join(a, '..', 'axe.js');
```
stephenmathieson, line 84: I have never understood why we didn't just do `require.resolve('axe-core')` here, but I'm 100% sure someone said there was a reason for it.

packages/cli/src/lib/utils.ts
```diff
@@ -71,34 +71,22 @@ export const parseBrowser = (browser?: string): string | Error => {
...
-      __dirname,
-      '..',
-      '..',
-      '..',
-      'node_modules',
-      'axe-core',
-      'axe.js'
-    );
+    /* Obtain the axe-core source from the globally installed CLI package instead */
+    axePath = path.join(__dirname, '..', 'node_modules', 'axe-core', 'axe.js');
```
stephenmathieson, line 89: Please explain why `require.resolve()` won't work.

Zidious: After looking into it again - we can rely on `require.resolve(..)` we execute a specific file from `dist`: `"axe": "./dist/cli.js"` - We don't _need_ to have ESM here - reverting back to use `require.resolve`

packages/cli/src/lib/utils.ts
```diff
@@ -76,18 +76,8 @@ export const getAxeSource = (axePath?: string): string | void => {
     return;
```
stephenmathieson, line 76: I'm confused by what's happening here.

If the file `axePath` exists, return nothing. Otherwise, return the contents of a file.

stephenmathieson (commented):
Does this need to be more complicated than:

```js
export const getAxeSource = (axePath = require.resolve('axe-core')): string => {
  return fs.readFileSync(axePath, 'utf-8')
}
```

...and if not, do we really need this function?

packages/cli/src/lib/utils.ts
Zidious: Refactored this ☝️

stephenmathieson (commented):
LGTM, but as [said before](https://github.com/dequelabs/axe-core-npm/pull/730#discussion_r1197003987), there was _probably_ a reason we did funky path lookups. Please make sure that there isn't some obscure requirement we're breaking by using `require.resolve()`.

stephenmathieson:
Should we remove the ESM stuff in the CLI package (since it's not used/needed)?

michael-siek:
I think we should not support ESM in CLI

Zidious:
> I think we should not support ESM in CLI

I am in favour of that - but IMO it should happen outside of this PR

stephenmathieson (approved):

PR #731, opened 2023-05-25
fix(playwright): use @playwright/test

dargmuesli:
Resolves the error `Please install @playwright/test package to use Playwright Test.`.
See https://github.com/microsoft/playwright/issues/23098.

Also removes a few unused variables.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -200,7 +200,7 @@ jobs:
 
   playwright:
     docker:
-      - image: mcr.microsoft.com/playwright:v1.29.0
+      - image: mcr.microsoft.com/playwright:v1.34.3
     environment:
       NODE_ENV: development
     working_directory: ~/axe-core-npm
--- a/packages/playwright/package-lock.json
+++ b/packages/playwright/package-lock.json
@@ -30,7 +30,7 @@
         "typescript": "^4.8.4"
       },
       "peerDependencies": {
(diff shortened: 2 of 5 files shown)


stephenmathieson:
@dargmuesli I was dealing with this issue in a different repository yesterday. I completely understand your pain.

However, I don't think having a peer dep on `@playwright/test` is the correct solution. Users may not be using the package, so we don't want to force them to have it installed. Can we simply have a peer dependency on `playwright-core` instead?

dargmuesli:
Thank you for the feedback! :heart:
I'd need to test that, I think I'll get around to do this either today or within this week.
Feel free to tinker with my branch though if you have quick ideas!

dargmuesli:
Using `playwright-core` as peer dependency worked as well for me! :) I've updated the code.

I think why the ci pipeline fails is not an error on my part, is it? Seems it cannot extract the downloaded browser image.

stephenmathieson (approved):
LGTM, but let's let a maintainer give the final OK.

----

Got the OK from @michael-siek 👍

stephenmathieson:
Reviewed for security

Zidious:
- [ ] Reviewed for security

stephenmathieson:
Thanks @dargmuesli! I'm not sure when the next public/stable release is planned for. In the meantime, you can `npm install @axe-core/playwright@next` in your project.

PR #735, opened 2023-06-01
ci: remove tests from the `develop` and `master` branch

Zidious:
I noticed we do not depend on the CircleCI playwright job for canary / prod releases. if it fails, we will still publish.

image

We don't need to run tests again on the develop and master branch as we run tests on each PR

Closes: https://github.com/dequelabs/axe-core-npm/issues/739

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -308,14 +308,6 @@ workflows:
           context: html-tools
           requires:
             - dependencies
-            - lint
-            - cli
-            - puppeteer
-            - webdriverjs
-            - webdriverio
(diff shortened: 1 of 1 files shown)


stephenmathieson (commented):
If the tests pass on `develop`, why do they need to be re-run on released branches?

Zidious:
> If the tests pass on `develop`, why do they need to be re-run on released branches?

Fair point - we could 100% save on minutes here as tests run on the PR, we do not need to run on dev branch too (will update)

stephenmathieson (approved):

PR #738, opened 2023-06-01
chore: better test structure / name conistency

straker:
Our test architecture was all over the place, so made it consistent between packages. The `tests` dir is now named `test` and the test files end in `.spec` instead of `.test` (in some). Also `esmTest.mjs` file got moved to inside the test dir.


packages/webdriverio/test/axe-webdriverio.spec.ts
```diff
@@ -67,7 +67,7 @@ describe('@axe-core/webdriverio', () => {
       });
     }
 
-    describe('WebdriverIO Async', () => {
+    describe(`WebdriverIO Async (${protocol} protocol)`, () => {
```
straker, line 70: This was needed so the describe wasn't the same

Zidious (approved):
LGTM

PR #741, opened 2023-06-01
docs: update docs to use named import and create example.js

straker:
Also made the docs consistent between packages (same url, same structure, etc.) and made the tests use the named import

### DIFF
--- a/packages/playwright/README.md
+++ b/packages/playwright/README.md
@@ -41,7 +41,7 @@ This module uses a chainable API to assist in injecting, configuring, and analyz
 Here is an example of a script that will drive Playwright to a page, perform an analysis, and then log results to the console.
 
 ```js
-const AxeBuilder = require('@axe-core/playwright').default;
+const { AxeBuilder } = require('@axe-core/playwright');
 const playwright = require('playwright');
 
 (async () => {
--- a/packages/playwright/README.md
+++ b/packages/playwright/README.md
@@ -6,33 +6,9 @@
 
 Install [Node.js](https://docs.npmjs.com/getting-started/installing-node) if you haven't already.
 
-Install Playwright:
+Install Playwright: `npm install playwright`
 
-NPM:
-
-```console
-npm install playwright
-```
-
-Yarn:
-
-```console
-yarn add playwright
-```
-
-Install `@axe-core/playwright` and its dependencies:
-
-NPM:
-
-```console
-npm install @axe-core/playwright
-```
-
-Yarn:
-
-```console
-yarn add @axe-core/playwright
-```
+Install @axe-core/playwright: `npm install @axe-core/playwright`
 
 ## Usage
 
@@ -45,7 +21,7 @@ const { AxeBuilder } = require('@axe-core/playwright');
 const playwright = require('playwright');
 
 (async () => {
-  const browser = await playwright.chromium.launch();
+  const browser = await playwright.chromium.launch({ headless: true });
   const context = await browser.newContext();
   const page = await context.newPage();
   await page.goto('https://dequeuniversity.com/demo/mars/');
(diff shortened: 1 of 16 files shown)


packages/puppeteer/README.md
```diff
@@ -25,12 +25,14 @@ const puppeteer = require('puppeteer');
 (async () => {
   const browser = await puppeteer.launch();
   const page = await browser.newPage();
-  await page.setBypassCSP(true);
```
straker, line 28: Not sure why we'd want this as part of our docs so removed it

packages/webdriverio/README.md
```diff
@@ -33,26 +33,29 @@ This module uses a chainable API to assist in injecting, configuring, and analyz
...
 ```js
-const AxeBuilder = require('@axe-core/webdriverio').default;
+const { AxeBuilder } = require('@axe-core/webdriverio');
 const { remote } = require('webdriverio');
 
 (async () => {
   const client = await remote({
     logLevel: 'error',
     capabilities: {
-      browserName: 'firefox'
```
straker, line 43: Firefox requires a binary that isn't installed by default on macs, so switched to chrome

packages/webdriverjs/src/index.ts
```diff
@@ -267,8 +267,4 @@ class AxeBuilder {
   }
 }
 
-if (typeof module === 'object') {
-  exports = module.exports = AxeBuilder;
-}
```
straker, line 272: I don't think this is necessary anymore so removed it

packages/puppeteer/example.js
```diff
@@ -0,0 +1,18 @@
...
+  await page.goto('https://dequeuniversity.com/demo/mars/');
+
+  try {
+    const results = await new AxePuppeteer(page).analyze();
+    console.log(results);
+  } catch (e) {
+    console.error(e);
+  }
+
+  await page.close();
```
Zidious, line 16: NDB - closing the browser should close all open pages
```suggestion

```

packages/playwright/example.js
```diff
@@ -1,12 +1,17 @@
-const AxeBuilder = require('@axe-core/playwright').default;
+const { AxeBuilder } = require('@axe-core/playwright');
 const playwright = require('playwright');
```
Zidious, line 2: playwright is outdated (I know this was already here) we should be using playwright-core: 

```suggestion
const playwright = require('playwright-core');
```

packages/playwright/example.js
```diff
@@ -1,12 +1,17 @@
...
 
 (async () => {
   const browser = await playwright.chromium.launch();
```
Zidious, line 5: As the other examples are in headless we should also add headless here too: 

```suggestion
  const browser = await playwright.chromium.launch({ headless: true });
```

Zidious (dismissed):
LGTM - left a few comments inline nothing blocking though

packages/playwright/README.md
```diff
@@ -6,33 +6,9 @@
 
 Install [Node.js](https://docs.npmjs.com/getting-started/installing-node) if you haven't already.
 
-Install Playwright:
+Install Playwright: `npm install playwright`
 
-NPM:
```
straker, line 11: No need to give yarn and npm examples so made this the consistent style for all the docs.

packages/puppeteer/README.md
```diff
@@ -8,9 +8,9 @@ Previous versions of this program were maintained at [dequelabs/axe-puppeteer](h
 
 Install [Node.js](https://docs.npmjs.com/getting-started/installing-node) if you haven't already.
 
-Install Puppeteer: `npm install puppeteer --no-save`
```
straker, line 11: Not sure why we are promoting `--no-save` option here when it's a required dependency to run

Zidious (approved):

PR #743, opened 2023-06-07
fix(webdriverjs): skip unloaded iframes

straker:
Closes https://github.com/dequelabs/axe-core-npm/issues/744

### DIFF
--- a/packages/webdriverjs/src/index.ts
+++ b/packages/webdriverjs/src/index.ts
@@ -1,4 +1,4 @@
-import { WebDriver } from 'selenium-webdriver';
+import { WebDriver, WebElement } from 'selenium-webdriver';
 import axe, {
   RunOptions,
   Spec,
@@ -30,6 +30,7 @@ export default class AxeBuilder {
   private builderOptions: BuilderOptions;
   private legacyMode = false;
   private errorUrl: string;
+  private frameStack: WebElement[]
 
   constructor(
     driver: WebDriver,
@@ -45,6 +46,7 @@ export default class AxeBuilder {
     this.builderOptions = builderOptions || {};
     this.errorUrl =
       'https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverjs/error-handling.md';
+    this.frameStack = [];
   }
 
   /**
@@ -174,7 +176,15 @@ export default class AxeBuilder {
       return this.runLegacy(context);
     }
 
-    const partials = await this.runPartialRecursive(context, true);
+    // ensure we fail quickly if an iframe cannot be loaded (instead of waiting
+    // the default length of 30 seconds)
+    const { pageLoad } = await this.driver.manage().getTimeouts();
+    this.driver.manage().setTimeouts({ pageLoad: 1000 });
+    this.frameStack = [];
+
+    const partials = await this.runPartialRecursive(context, null, true);
+
+    this.driver.manage().setTimeouts({ pageLoad });
 
     try {
       return await this.finishRun(partials);
@@ -212,6 +222,7 @@ export default class AxeBuilder {
    */
   private async runPartialRecursive(
     context: SerialContextObject,
+    frame: null | WebElement,
     initiator = false
   ): Promise<string[]> {
     if (!initiator) {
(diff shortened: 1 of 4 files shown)


packages/webdriverjs/src/index.ts
```diff
@@ -174,7 +176,15 @@ export default class AxeBuilder {
...
-    const partials = await this.runPartialRecursive(context, true);
+    // ensure we fail quickly if an iframe cannot be loaded (instead of waiting
+    // the default length of 30 seconds)
+    const { pageLoad } = await this.driver.manage().getTimeouts();
+    this.driver.manage().setTimeouts({ pageLoad: 1000 });
+    this.frameStack = [];
+
+    const partials = await this.runPartialRecursive(context, null, true);
+
+    this.driver.manage().setTimeouts({ pageLoad });
```
dbjorge, line 187: Consider wrapping L185 in a try/finally and put restoring the user's driver state in a finally block? Would be very confusing for a user if an error in one axe scan caused subtle timeout behavior changes in the user's subsequent tests

packages/webdriverjs/src/index.ts
```diff
@@ -223,18 +234,31 @@ export default class AxeBuilder {
...
+        /*
+          When switchTo().frame() fails using chromedriver (but not gekodriver)
+          it puts the driver into a really bad state that is impossible to
+          recover from. So we need to switch back to the main window and then
+          go back to the desired iframe context
+        */
+        const win = await this.driver.getWindowHandle();
+        await this.driver.switchTo().window(win);
+
+        for (const frameElm of this.frameStack) {
```
dbjorge, line 258: Because the recursive call can manipulate `this.frameStack`, it feels a bit tricky to be confident that this is guaranteed to always restore to the frame stack as it existed without recursive calls, regardless of if/how any errors happened within the recursive call.

Maybe preferable would be to avoid using a mutable property for `this.frameStack` at all, and instead pass the `frameStack` (as `[...frameStack, frame]`) rather than the `frame` to `runPartialRecursive`? Then it would be a lot easier to be confident that this catch handler is restoring to the expected pre-state regardless of if/how any inner `runPartialRecursive` call might fail.

packages/webdriverjs/src/index.ts
```diff
@@ -1,4 +1,4 @@
-import { WebDriver } from 'selenium-webdriver';
+import { WebDriver, WebElement } from 'selenium-webdriver';
```
Zidious, line 1:
```suggestion
import type { WebDriver, WebElement } from 'selenium-webdriver';
```

packages/webdriverjs/src/index.ts
```diff
@@ -224,17 +234,25 @@ export default class AxeBuilder {
...
         await this.driver.switchTo().parentFrame();
       } catch {
-        if (switchedFrame) {
-          await this.driver.switchTo().parentFrame();
+        /*
+          When switchTo().frame() fails using chromedriver (but not geckodriver)
+          it puts the driver into a really bad state that is impossible to
+          recover from. So we need to switch back to the main window and then
+          go back to the desired iframe context
+        */
```
Zidious, line 248: 👍

packages/webdriverjs/test/fixtures/iframe-lazy-1.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
```
Zidious, line 1: NBD - Do we want these fixtures to live in the axe-fixtures repo?

Zidious (changes requested):
Small things inline -

packages/webdriverjs/test/fixtures/iframe-lazy-1.html
straker: Good question. I should probably move them since puppeteer also has this same issue. I'll add this as a tech debt ticket since I'd like to get this released without waiting to add it to the fixture repo.

straker: https://github.com/dequelabs/axe-test-fixtures/issues/22

Zidious (approved):

Issue #727, opened 2023-05-10
Puppeteer hangs when page has lazy loaded iframe


straker:
When a page has an iframe with `loading="lazy"` attribute and that has not been scrolled into view,  Puppeteer [will hang indefinitely if you try to interact with a non-loaded iframe](https://github.com/puppeteer/puppeteer/issues/8399) (e.g. using `frame.$$(selector)` never throws nor completes). What that means is if you try to scan a page with such an iframe the analyze will never finish (e.g. https://www.rastoder.si/).

We should inform the user that the iframe isn't loaded and skip it like we do when we can't inject axe into the iframe.

PR #752, opened 2023-06-14
fix: skip unloaded iframes for all apis

straker:
Implements the webdriverjs fix into all apis. Playwright was the only api that could handle unloaded iframes without issues, so it didn't need to do anything special and thus can give different results than the other apis if there are unloaded iframes on the page.

This fix doesn't fix the issue for legacy mode. If we want to support legacy mode and unloaded iframes we should do it in a separate pr.

Closes https://github.com/dequelabs/axe-core-npm/issues/727


packages/webdriverio/test/axe-webdriverio.spec.ts
```diff
@@ -850,6 +850,40 @@ describe('@axe-core/webdriverio', () => {
...
+          it('resets pageLoad timeout to user setting', async () => {
+            await client.url(`${addr}/lazy-loaded-iframe.html`);
+            client.setTimeout({ pageLoad: 500 })
+            const title = await client.getTitle();
+
+            const results = await new AxeBuilder({client})
+              .options({ runOnly: ['label', 'frame-tested'] })
+              .analyze();
+
+            const timeout = await client.getTimeouts();
```
WilcoFiers, line 884: Do you need to reset this in an after call?

WilcoFiers (dismissed):
Couple small points

packages/puppeteer/src/utils.ts
```diff
@@ -76,7 +76,16 @@ export async function assertFrameReady(frame: Frame): Promise<void> {
   // Check if the page is loaded.
   let pageReady = false;
   try {
-    pageReady = await frame.evaluate(pageIsLoaded);
+    // Puppeteer freezes on unloaded iframes. Set a race timeout in order to handle that.
+    // @see https://github.com/dequelabs/axe-core-npm/issues/727
+    const timeoutPromise = new Promise(resolve => {
+      setTimeout(() => {
+        resolve('timeout');
```
stephenmathieson, line 83: Resolving with the value "timeout" is not at all idiomatic. If we time out, that is a sign of a problem, which is what the `reject` callback is meant to be used for. If we reject this promise (as discussed), we'd enter the `catch` block below (which _should_ throw the `Page/Frame is not ready` error).

stephenmathieson (dismissed):
Let's reject promises rather than fulfilling them with an error.

packages/puppeteer/src/utils.ts
dylanb: we do not want to throw in these scenarios

packages/webdriverio/src/utils.ts
```diff
@@ -98,6 +101,39 @@ export const axeSourceInject = async (
...
+      client.execute() on an unloaded iframe would cause
+      the code to hang indefinitely since it is using
+      Puppeteer which freezes on unloaded iframes. Set a
+      race timeout in order to handle that. Code taken
+      from our @axe-core/puppeteer utils function.
+      @see https://github.com/dequelabs/axe-core-npm/issues/727
+    */
+    const timeoutPromise = new Promise(resolve => {
+      setTimeout(() => {
+        resolve('timeout');
```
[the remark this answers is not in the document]
dylanb: we really don't want to fail in these scenarios, we want to analyze the page content that we can analyze

packages/puppeteer/src/utils.ts
stephenmathieson: We do throw in these scenarios. See below.

packages/webdriverio/src/utils.ts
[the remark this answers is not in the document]
stephenmathieson: We use the error as flow control. See below.

packages/webdriverio/test/axe-webdriverio.spec.ts
straker: Good point, this is bleeding through the tests which could cause problems.

packages/webdriverio/src/utils.ts
```diff
@@ -98,6 +101,33 @@ export const axeSourceInject = async (
...
+      setTimeout(() => {
+        reject();
+      }, FRAME_LOAD_TIMEOUT)
+    });
+    const executePromise = client.execute(() => {
+      return document.readyState === 'complete'
+    });
+    await Promise.race([timeoutPromise, executePromise]);
+  } catch {
+    throw new Error('Page/Frame is not ready');
```
dylanb, line 127: Isn't part of the point of making these changes that we do not throw an exception?

dylanb (changes requested):
What is the effect on an analyze call of throwing that exception? If it is going to cause the analyze call to throw, then I don't think that is the correct behavior

stephenmathieson:
LGTM but I'll wait to approve until Dylan's question is answered.

packages/webdriverio/src/utils.ts
straker: We decided that the quickest way to resolve the bug was to return these iframes as `frame-tested` issues. How we want to handle unloaded / lazy loaded iframes is still a topic for discussion and being decided, but that shouldn't slow us down getting a fix out to the client.

stephenmathieson (approved):

PR #759, opened 2023-07-07
ci: Replace CircleCI with GitHub Actions

stephenmathieson:
Also test Node.js versions 16, 18 and 20.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node}}
+      - uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      - run: npm run build --prefix=packages/puppeteer
+      # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
+      - shell: 'script -q -e -c "bash {0}"'
```
straker, line 38: Should we use https://github.com/browser-actions/setup-chrome instead of relying on browser-driver-manager to install chrome? That would bypass the need for this hack to force TTY. Then browser-driver-manager could just install chromedriver which shouldn't need to manipulate the stdout.

stephenmathieson: Possibly, I can play around with that.

FWIW if we didn't have the `chromedriver` dev dep here, we wouldn't need to do anything (see https://github.com/dequelabs/jazzband/pull/486).

straker: Ya, the need to have the npm chromedriver match the installed version of chrome causes so much pain.

stephenmathieson: For jazzband, we just assume the developer (or CI container) has chromedriver on `$PATH`. Can we do the same here?

straker: No, because webdriverio has a direct dependency on the npm chromedriver [in order to setup the test](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/webdriverio/test/axe-webdriverio.spec.ts#L58)

straker: Wait, looks like if it's on path we don't need chromedriver, so maybe you're correct?

straker: CLI would at least [need it or pass in the path manually](https://github.com/dequelabs/axe-core-npm/blob/develop/packages/cli/src/lib/webdriver.ts#L14). Not sure if letting it be on `$PATH` there would be sufficient.

stephenmathieson: Using the `setup-chrome` option still requires us to have the correct `chromedriver` version installed. Since we're not going to remove the `chromedriver` dep (until we move over to `@puppeteer/browsers`), my vote is we keep this as-is.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - run: npm ci
```
straker, line 17: This should be [gated by a cache miss](https://github.com/actions/cache#example-cache-workflow) otherwise it will always run and invalidate the cache.

```yml
      - name: Install dependencies
        if: steps.cache-deps.outputs.cache-hit != 'true'
        run: npm ci
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      - run: npm run build --prefix=packages/react
+      - run: npm run test --prefix=packages/react
+      - run: npm run test:esm --prefix=packages/react
+
+  playwright:
+    strategy:
+      matrix:
+        node: [16, 18, 20]
```
straker, line 153: How is this working when we said we had to [force pin to node 14](https://github.com/dequelabs/axe-core-npm/blob/ca07660635fbc6c0e498d4326213b3901bdf3698/.circleci/config.yml#L207) due to WDIO sync?

.github/workflows/tests.yml
stephenmathieson: Why/how would `npm ci` invalidate a cache?

.github/workflows/tests.yml
stephenmathieson: We don't support WDIO sync. See https://github.com/dequelabs/axe-core-npm/pull/708

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+    runs-on: ubuntu-latest
+    needs: lint
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+      - uses: actions/cache@v3
```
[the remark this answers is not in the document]
stephenmathieson: Yes, using `cache: 'npm'` would be much better here 👍

.github/workflows/tests.yml
straker: https://docs.npmjs.com/cli/v9/commands/npm-ci

> If a node_modules is already present, it will be automatically removed before npm ci begins its install

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
+name: Deploy
+
+on:
+  push:
+    branches:
+      - develop
+      - master
+
+jobs:
```
dbjorge, line 9: These should both be gated on something like `if: github.repository_owner == 'dequelabs'` to avoid the possibility of forks accidentally trying to run them

.github/workflows/tests.yml
straker: Oh right, forgot we finally resolved that issue.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      # Make sure the working tree is clean
+      - run: git reset --hard HEAD
+      - run: |
+          npx lerna publish from-package \
+            --force-publish='*' \
+            --yes
+      - uses: softprops/action-gh-release@v1
```
dbjorge, line 53: Tags are mutable, so this sort of `uses` directive amounts to using an unpinned dependency version in an environment containing a publishing secret. Safer to pin to a commit hash rather than the `v1` tag.

stephenmathieson: While it's safer, it's also a maintenance nightmare ☹️

dbjorge: Renovate supports a syntax to [use digest pinning but also understand to update according to tags](https://docs.renovatebot.com/modules/manager/github-actions/#additional-information), which is the best of both worlds. That's what I recommend we end up with.

dbjorge: Dependabot [also appears to support this](https://github.com/dependabot/dependabot-core/pull/5951), though I didn't see documentation about it.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,156 @@
...
+        node: [16, 18, 20]
+    runs-on: ubuntu-latest
+    needs: lint
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node}}
+          cache: 'npm'
+      - run: npm run build --prefix=packages/puppeteer
```
dbjorge, line 30: Most jobs below this line are missing an `npm ci` step after `actions/setup-node`

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+jobs:
+  canary:
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'develop'
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - uses: actions/cache@v3
```
dbjorge, line 18: Did you mean to leave these two using `actions/cache` instead of `setup-node` caching + `npm ci`?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,156 @@
...
+        node: [16, 18, 20]
+    runs-on: ubuntu-latest
+    needs: lint
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+          cache: 'npm'
+      - run: npx playwright install --with-deps
```
dbjorge, line 141: If this job is a long-pole, we might consider running it on a playwright docker image pinned to the same version as our playwright version instead of `ubuntu-latest` - they come with playwright deps + matching browser versions pre-installed

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      # Make sure the working tree is clean
+      - run: git reset --hard HEAD
+      - run: |
+          npx lerna publish prepatch \
```
dbjorge, line 25: Do these commands actually have permissions to publish to NPM? I'd have expected to need to propagate an NPM_TOKEN secret or similar

dbjorge: (and probably a GH_TOKEN with extended permissions for `action-gh-release`)

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,53 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
+      - uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      # Make sure the working tree is clean
+      - run: git reset --hard HEAD
```
dbjorge, line 23: Was this here previously for some reason that was specific to circleci? This feels like something that should be a "fail the build if not true" step, not a "silently force it to be true" step

dbjorge (changes requested):
Thanks Stephen! Suggestions inline

.github/workflows/tests.yml
```diff
@@ -0,0 +1,185 @@
...
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node }}
+      - uses: actions/cache@v3
+        with:
+          path: '**/node_modules'
+          key: ${{ runner.os }}-npm-${{ hashFiles('**/package-lock.json') }}
+      - run: npx playwright install --with-deps
+      - run: npm ci
```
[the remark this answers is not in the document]
stephenmathieson: Yes

.github/workflows/tests.yml
stephenmathieson: We're now caching `~/.npm` so this is no longer a concern.

.github/workflows/deploy.yml
stephenmathieson: The `action-gh-release` uses the GHA token by default. I added a secret (`NPM_TOKEN`) and am now using it.

.github/workflows/tests.yml
stephenmathieson: Good idea! This job does take significantly longer than similar jobs.

playwright vs puppeteer

Are you OK with leaving this as-is and using the Playwright Docker image in a followup PR? I can likely get to it within the next few days.

.github/workflows/deploy.yml
stephenmathieson: I'll pin this action to a commit, but am worried that unless we get Renovate setup, using actions this way is going to be a pain.

dbjorge (approved):

PR #761, opened 2023-07-07
chore: Make existing ESLint warnings fail the build

stephenmathieson:
People don't care about warnings. Unless a build fails, ESLint output will go unnoticed. This patch prevents that from happening by converting existing warnings to errors.

Also all existing warnings have been fixed.

### DIFF
--- a/packages/webdriverio/src/utils.ts
+++ b/packages/webdriverio/src/utils.ts
@@ -10,7 +10,7 @@ import type {
   SerialContextObject
 } from 'axe-core';
 
-export const FRAME_LOAD_TIMEOUT = 1000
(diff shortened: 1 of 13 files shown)


packages/webdriverio/src/utils.ts
```diff
@@ -152,6 +152,8 @@ export const axeRunPartial = (
 export const axeGetFrameContext = (
   client: Browser,
   context: SerialContextObject
+  // TODO: add proper types
```
straker, line 155: Was this left in intentionally?

stephenmathieson: Yes

straker (approved):

PR #767, opened 2023-07-12
test: Make axe-core version test more robust

stephenmathieson:
- Test every package (under `packages/*`)
- Use the battle-tested [`semver`](https://www.npmjs.com/package/semver) library rather than a regex


WilcoFiers (approved):
Code looks fine, not sure why CI is failing though.

PR #786, opened 2023-08-08
ci: add default timeout to test suite

Zidious:
We shouldn't let tests run for large amounts of time: https://github.com/dequelabs/axe-core-npm/actions/runs/5790182483/job/15692791377 in case of hanging.

Set a default timeout for linting at 5min and integration tests at 15mins.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,6 +7,10 @@ on:
       - "!master" # except master
       - "!develop" # except develop
 
+env:
+  # Point tests to run from newest stable chromedriver
+  CHROMEDRIVER_PATH: /usr/local/bin/chromedriver
+
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,20 +1,11 @@
 name: Tests
 
-on:
-  push:
-    branches:
-      - "**" # all branches
-      - "!master" # except master
-      - "!develop" # except develop
-
-env:
-  # Point tests to run from newest stable chromedriver
-  CHROMEDRIVER_PATH: /usr/local/bin/chromedriver
+on: [push]
 
 jobs:
   lint:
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yml
```diff
@@ -1,10 +1,20 @@
...
+      - "!develop" # except develop
+
+env:
+  # Point tests to run from newest stable chromedriver
+  CHROMEDRIVER_PATH: /usr/local/bin/chromedriver
 
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
```
dbjorge, line 17: This is only adding a timeout to the lint job specifically, but the PR title/description makes me think you intended to add one to all jobs, no?

(also, 30 minutes is probably excessive - I'd suggest something more like 5m for the lint job and 15m for the browser-based e2e jobs)

.github/workflows/tests.yml
```diff
@@ -1,10 +1,20 @@
 name: Tests
 
-on: [push]
+on:
+  push:
+    branches:
+      - "**" # all branches
+      - "!master" # except master
+      - "!develop" # except develop
```
dbjorge, line 8: I think it's important that we run checks against these branches because they enable tests to catch issues with logical merge conflicts. Particularly, `on: push` for a PR is generally going to run checks against the commits in the PR, not the merge commit for that PR back into its target branch; including `develop` ensures that we get test coverage of the merged result.

We could consider using `on: pull_request` to cover PRs instead (which runs against merge commits rather than direct commits), but it has the disadvantage that you then don't start running tests until after you've finished writing up a PR description, and it's nice in practice to be able to start seeing test results earlier.

straker (approved):
It's disappointing that you have to manually set the timeout per job. Hopefully someday they'll allowing [setting the timeout globally](https://github.com/orgs/community/discussions/10690).

Zidious:
superseded by: https://github.com/dequelabs/axe-core-npm/pull/795

PR #789, opened 2023-08-09
test: fix node v20 tests

AdnoC:


packages/cli/src/lib/utils.test.ts
```diff
@@ -73,25 +75,33 @@ describe('utils', () => {
...
+        dirname = join(tempDir, 'packages', 'cli', 'lib');
+        mkdirSync(dirname);
       });
 
-      afterEach(() => {
-        mock.restore();
-      });
       it('fall back to use `locally` installed axe-core', () => {
-        const axeSource = utils.getAxeSource();
+        const axeSource = utils.getAxeSource(undefined, dirname, dirname);
```
dbjorge, line 99: With this update, I feel a bit confused about what this test is trying to actually verify - I'm not sure there's any meaningful real case where `cwd` === `dirname` in practice. I think it would be preferable to break this out into separate test cases for:

* `it('uses axe.js from the working directory if it exists')` (where test setup places `axe.js` in all 3 locations with different version numbers, to verify which version is actually found)
* `it("falls back to axe-core from the working directory's node_modules if axe.js doesn't exist in the working directory")` (where test setup places `axe.js` in both the dirname and cwd based node_modules paths, but not in cwd, again with different versions to verify which one is found)
* `it("falls back to axe-core from our own package's node_modules if no working-directory based implementation exists")` (where test setup only populates the dirname based node_modules path)

packages/cli/src/lib/utils.test.ts
```diff
@@ -73,25 +75,33 @@ describe('utils', () => {
...
   describe('getAxeSource', () => {
-    describe('mock file', () => {
-      beforeEach(() => {
-        mock({
-          '/node_modules/axe-core': {},
-          '../node_modules/axe-core': {
-            'axe.js': mock.load(require.resolve('axe-core'))
-          }
-        });
+    describe.only('mock file', () => {
```
dbjorge, line 78: Stray `.only`

packages/cli/package-lock.json
```diff
@@ -53,9 +52,9 @@
       }
     },
     "node_modules/@axe-core/webdriverjs": {
-      "version": "4.7.1",
-      "resolved": "https://registry.npmjs.org/@axe-core/webdriverjs/-/webdriverjs-4.7.1.tgz",
-      "integrity": "sha512-Cf3gFDKYkMDo1vO9QQ3uIRP/iyiacxbu6YWFl0LqKG3u4AtUeR64xZ5weexr5xAKixu/sDQmcpjLskkgO9QK9w==",
+      "version": "4.7.3",
```
dbjorge, line 55: This is an example of the type of stray `package-lock.json` update I mentioned finding concerning in standup earlier - I can live with including it here, just noting it for ease of referencing in retro tomorrow

dbjorge (dismissed):

packages/cli/src/lib/utils.test.ts
dbjorge: (the `axe-core` repo uses `eslint-plugin-mocha-no-only` to prevent this class of mistake from being merged accidentally, maybe would be good for a separate PR to add that in this repo as well)

straker (dismissed):

packages/cli/src/lib/utils.test.ts
```diff
@@ -74,24 +76,58 @@ describe('utils', () => {
...
+        assert.include(axeSource, 'cwd');
       });
-      it('fall back to use `locally` installed axe-core', () => {
-        const axeSource = utils.getAxeSource();
-        const axeVersionCheck = dependencies['axe-core'].replace('^', '');
-        assert.include(axeSource, axeVersionCheck);
+      it("falls back to axe-core from the working directory's node_modules if axe.js doesn't exist in the working directory", () => {
+        const { cliDirname, cwdDirname } = setupTree();
+        rmSync(join(cwdDirname, 'axe.js'));
+        const axeSource = utils.getAxeSource(undefined, cliDirname);
```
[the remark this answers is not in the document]
AdnoC: Nope. Should be `cliDirname` because we need a directory with `node_modules` in it.

straker (approved):

PR #797, opened 2023-08-22
ci: fix deploy sha

straker:
Lerna was generating SHAs that weren't compatible with npm tags. E.g `@axe-core/cli@4.7.4-alpha.0+8b5d296` which would be truncated to `@axe-core/cli@4.7.4-alpha.0` for npm.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -29,6 +29,7 @@ jobs:
             --canary \
             --dist-tag=next \
             --exact \
+            --preid=$(git rev-parse --short "$GITHUB_SHA") \
             --force-publish \
             --yes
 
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -29,6 +29,7 @@ jobs:
             --canary \
             --dist-tag=next \
             --exact \
+            --preid=$(git rev-parse --short "$GITHUB_SHA") \
```
stephenmathieson, line 32: Why use `$GITHUB_SHA` rather than `HEAD`?

stephenmathieson (dismissed):

.github/workflows/deploy.yml
straker: What is `HEAD`? It's not listed in https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables

stephenmathieson: It's the "head" of the current branch.

stephenmathieson (approved):

Issue #802, opened 2023-08-23
Request: Export the default `logToConsole` logger


hiddenist:
I'm looking into adding some UI when accessibility violations are found while using `@axe-core/react`, but I still want to have the default console logging running. This would be fairly straightforward if the `logToConsole` function was exported.

https://github.com/dequelabs/axe-core-npm/blob/acf6b0fcbae2dc7069a15b17871a64dde66191b9/packages/react/index.ts#L314

michael-siek:
Hey @hiddenist, I will bring this up with the team and will update you soon!

michael-siek:
Hey @hiddenist 
Please feel free to follow this PR https://github.com/dequelabs/axe-core-npm/pull/812 for exporting `logToConsole`

padmavemulapati:
This is not actually an issue, but the way results rendered to console to specific requirement. i.e.,"Request: Export the default logToConsole logger" so that it cannt be considered as a feature, which can implement all over the packages.

It can be applicable only for the react-componants, those can be rendered to the browsers and browser's devtools console would be showing the logs for `New Issues`

Steps validated:

1. create a component 
```index.html
<!DOCTYPE html>
<html>
  <head> </head>
  <body>
    <div id="container"></div>
    <img src=""/>
    <style>
      * {
        padding: 0;
        margin: 0;
      }

      html {
        font: 14px normal Arial, sans-serif;
        color: #626771;
        background-color: #fff;
      }

      body {
        padding: 60px;
        text-align: center;
      }

      h1 {
        font-size: 18px;
        margin-bottom: 20px;
      }
    </style>
    <script src="dist/bundle.js"></script>
  </body>
</html>
```
2. configure the axe to the componant

```
const axeConf = {
  rules: [
    { id: 'heading-order', enabled: true },
    { id: 'label-title-only', enabled: true },
    { id: 'image-alt', enabled: true },
    { id: 'link-in-text-block', enabled: true },
    { id: 'region', enabled: true },
    { id: 'skip-link', enabled: true },
    { id: 'help-same-as-label', enabled: true }
  ]
};
```
3. Render the above react-componant via browser (here started the server on http://localhost:8081/) using 

```
if (process.env.NODE_ENV !== 'production') {
  axe(React, ReactDOM, 1000, axeConf);
  const axeResults = {} // get axe-results from somewhere

   axe.logToConsole(axeResults) // logs to the devtools console
}
```
[...]

PR #812, opened 2023-08-31
feat(react): export `logToConsole` fn

michael-siek:
closes: https://github.com/dequelabs/axe-core-npm/issues/802


straker (changes requested):
Please add a test to show that this is indeed exported.

straker (approved):

PR #831, opened 2023-09-22
Export AxeResults type from `@axe-core/playwright` package

jonchurch:
We had a need to get the `AxeResults` type, to pass to [`axe-sarif-converter`](https://www.npmjs.com/package/axe-sarif-converter) and we realized there is no good way to get the type without importing `axe-core` itself, or doing type gymnastics like:

```
import AxeBuilder from '@axe-core/playwright';
type AxeResultsType = ReturnType<AxeBuilder['analyze']> extends Promise<infer T> ? T : never;
```

Would you consider this PR to export AxeResults from the playwright package?

> For anyone who finds this, another approach, if specifically using `axe-sarif-converter` is:
> `type AxeResultsType = Parameters<typeof convertAxeToSarif>[0];`

### DIFF
--- a/packages/playwright/src/index.ts
+++ b/packages/playwright/src/index.ts
@@ -327,3 +327,4 @@ export default class AxeBuilder {
 }
 
 export { AxeBuilder };
+export type { AxeResults }


straker:
Thanks for bringing this to our attention. We've decided that we will export the entire `axe` type in all our apis as the `AxeResults` type uses many sub types that will also be helpful/necessary to have available. We'll open a pr soon that does this to all the apis as once. As such we'll close this one. Thanks again.

PR #844, opened 2023-09-28
chore: update to use `lerna@7`

michael-siek:


lerna.json
```diff
@@ -1,4 +1,3 @@
 {
-  "useWorkspaces": true,
```
michael-siek, line 2: Lerna v7 no longer uses `useWorkspaces` option

straker (approved):

straker (dismissed):

straker (approved):

PR #854, opened 2023-10-04
ci: ensure packages are built

michael-siek:


package.json
```diff
@@ -11,7 +11,8 @@
     "fmt": "prettier --write .",
     "lint": "eslint .",
     "prepare": "husky install",
-    "precommit": "lint-staged"
+    "precommit": "lint-staged",
+    "build": "lerna run build"
```
Zidious, line 15: We're using npm workspaces here, we can utilise that to build all packages instead of lerna:
```suggestion
    "build": "npm run --workspaces build"
```

.github/workflows/deploy.yml
```diff
@@ -17,6 +17,8 @@ jobs:
           node-version: 18
           cache: 'npm'
       - run: npm ci
+      # Ensure packages are built
+      - run: npm run build
```
Zidious, line 21: I suspect we'll need this for production releases too

Zidious (dismissed):
Two minor comments inline -

Zidious (approved):

Issue #849, opened 2023-10-02
Create GHA for manually triggering create-release-candidate


dequejenn:
Create a GHA that calls the public action `@axe-api-team-public/create-release-candidate`. For now, don't schedule it, just add a manual trigger that has a checkbox and set variables as follows:

```
name: Create release candidate 

on:
  workflow_dispatch
    inputs:
      version-locked:
        type: boolean
        description: Follow semver versioning if not version locked
        default: true
jobs:
  create-release-candidate:
    runs-on: ubuntu-latest
    steps:
      - uses: dequelabs/axe-api-team-public/.github/actions/create-release-candidate-v1@main
        with:
          token: ${{ secrets.GH_PROJECT_TOKEN }}
          slack-webhook: ${{ secrets.SLACK_WEBHOOK }}
          base: 'master'
          head: 'develop'
          release-script-path: '/.github/scripts/prepare_release.sh'
          version-locked: ${{ inputs.version-locked }}
        env:
          GH_TOKEN: ${{ secrets.GH_PROJECT_TOKEN }}
```

PR #864, opened 2023-10-11
ci: add `create-release-candidate.yml`

Zidious:
Closes: https://github.com/dequelabs/axe-core-npm/issues/849


.github/workflows/create-release-candidate.yml
```diff
@@ -0,0 +1,23 @@
+name: Create release candidate 
+
+on:
+  workflow_dispatch:
+    inputs:
+      version-locked:
+        type: boolean
+        description: Uncheck if you want to follow SemVer versioning rather than version-lock to axe-core.
```
straker, line 8: This isn't a great way to put this. When updating axe-core to a minor or major version it would be still version-locked to release as a major version but the release-candidate script itself would prevent that. Instead I propose this as the wording (@dequejenn should approve to):

> Should this action be allowed to create a minor or major release?

With the default being the opposite value of the `version-locked` setting of the repo.

dequejenn: @straker - what Gabe and I were realizing is if we had the checkbox be `checked=version locked` and `unchecked=not version locked/semver` then there wasn't additional code to map the input field. I am fine with changing the text but the test should lead someone to check the box only if we are NOT following semver.

dequejenn: @Zidious - Steve came up with the best one. Please change wording on teh checkbox to  `Prevent this action from creating a minor or major release`

straker: So the input will still be for the `version-locked` variable, it's just the wording change. For the desire to have check the checkbox box === not following semver then we can use the wording:

> Prevent this action from creating a minor or major release

straker (approved):

.github/workflows/create-release-candidate.yml
Zidious: Tested visuals in test-repo: 

image

Issue #867, opened 2023-10-11
Add creating a canary on push into release in deploy.yml 


dequejenn:
Currently we only create a canary build on commit into develop. However, there is a chance that the commits in develop were never pushed into the release branch for QA. To mitigate this use case, we need to also build a canary when a push to the release branch happens, so QA is always testing the correct codeset that will be deployed. 

In deploy.yaml,
```
on:
  push:
    branches:
      - develop
      - master
```
needs to be 
```
on:
  push:
    branches:
      - develop
      - release
      - master
```

AND

```
canary:
    runs-on: ubuntu-latest
    if: github.ref_name == 'develop' && github.repository_owner == 'dequelabs'
```
needs to be
```
canary:
    runs-on: ubuntu-latest
    if: (github.ref_name == 'develop' || github.ref_name == 'release') && github.repository_owner == 'dequelabs'
```

PR #875, opened 2023-10-13
ci: add `release-candidate` to deploy.yml

Zidious:
Closes: https://github.com/dequelabs/axe-core-npm/issues/867

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   canary:
     runs-on: ubuntu-latest
-    if: (github.ref_name == 'develop' || github.ref_name == 'release') && github.repository_owner == 'dequelabs'
+    if: github.ref_name == 'develop' && github.repository_owner == 'dequelabs'
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - develop
+      - release
```
straker, line 7: The release branch will have a date appended to it

```suggestion
      - release-*
```

.github/workflows/deploy.yml
```diff
@@ -34,6 +35,33 @@ jobs:
             --force-publish \
             --yes
 
+  relase-candidate:
+    runs-on: ubuntu-latest
+    if: github.ref_name == 'release' && github.repository_owner == 'dequelabs'
```
straker, line 40: With the date appended this will never return true. Will need to update to handle dates in the branch name

straker (dismissed):

.github/workflows/deploy.yml
Zidious: (discussed on slack)

straker (approved):

PR #880, opened 2023-10-18
refactor: use new test-fixtures path

straker:
Use the new exported `fixturesPath` from `axe-test-fixtures` and remove all the symlinks to `external` and all `fixtures` directories.


package.json
```diff
@@ -17,6 +17,7 @@
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "^5.40.0",
     "@typescript-eslint/parser": "^5.40.0",
+    "axe-test-fixtures": "github:dequelabs/axe-test-fixtures#v1",
```
straker, line 20: This wasn't pinned to `v1` so wasn't getting the path update. Also not sure why it was a dependency.

packages/webdriverjs/test/axe-webdriverjs.spec.ts
```diff
@@ -619,7 +616,7 @@ describe('@axe-core/webdriverjs', () => {
     });
 
     it('with chaining include and exclude', async () => {
-      await driver.get(`${addr}/context.html`);
+      await driver.get(`${addr}/context-include-exclude.html`);
```
WilcoFiers, line 619: This test doesn't really prove anything. If you only had the .includes the assertions would still pass because the excludes aren't nested in an include. Not sure that needs to be fixed in this PR, but something should be done about it IMO.

WilcoFiers (changes requested):
Small comment, please have a look

packages/webdriverjs/test/axe-webdriverjs.spec.ts
straker: Yup. Noticed that too. This would fix it https://github.com/dequelabs/axe-test-fixtures/pull/28

Zidious (approved):
LGTM

Issue #798, opened 2023-08-23
Error occurs when WebdriverIO version is higher than 5


abakarovmaks:
### Product

webdriverio

### Product Version

4.7.3

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
The test must pass succesfully.

#### Actual
Test fails due to error "An instantiated WebdriverIO client greater than v5 is required"

#### How to Reproduce
```
import { browser } from '@wdio/globals';
import AxeBuilder from '@axe-core/webdriverio';

async function checkAxeAccesability() {
  const builder = new AxeBuilder({ client: browser });
  const result = await builder.analyze();
  const violations = result.violations.filter(
    (violation) => violation.impact === 'critical'
  );
  const errorMessages = violations.map((violation, index) => {
    return `${index + 1}. ${violation.help}`;
  });
  const error = errorMessages.join('\n');
  assert.strictEqual(violations.length, 0, error);
}

await checkAxeAccesability();
```

#### Additional context
I trying to run the test in WebdriverIO.
My dependencies: 
```
    "@axe-core/webdriverio": "^4.7.3",
    "axe-core": "^4.7.2",
```
My WebdriverIO version is 8.14.6, Chromedriver version is  116.0.5845.96
When I run the test I get an error "An instantiated WebdriverIO client greater than v5 is required". (but my version of WebdriverIO is 8.14.6).
But when I remove some code from the file index.js which is in the folder node_modules/@axe-core/webdriverio/dist everything works fine.
Removed code from index.js:
```
  (0, import_assert2.default)(
      isWebdriverClient(client),
      "An instantiated WebdriverIO client greater than v5 is required"
    );
```
and
```
var isWebdriverClient = (client) => {
  if (!client || typeof client !== "object") {
    return false;
  }
  if (typeof client.execute !== "function") {
    return false;
  }
  if (typeof client.switchToFrame !== "function") {
    return false;
  }
  return true;
};
```
Is it the global issue with new version of WebdriverIO or what?

Thank you in advance for understanding.
Best regards.

padmavemulapati:
Verified with the latest axe-core/webdriverio - 4.7.4-acf6b0f.0

Able to see chromedriver - 116 version and test ran successfully
Steps followed:
1. npm init -y
2. npm install @axe-core/webdriverio@next (4.7.4-acf6b0f.0)
3. npm install webdriverio(8.16.4)
4. npm install chromedriver(116.0.0)
5. added test.js
[...]

PR #882, opened 2023-10-20
fix(webdriverio): work with @wdio/globals

straker:
When using `@wdio/globals`, `browser` is a function and not an object (thanks wdio).

Closes: https://github.com/dequelabs/axe-core-npm/issues/798

### DIFF
--- a/packages/webdriverio/src/utils.ts
+++ b/packages/webdriverio/src/utils.ts
@@ -16,7 +16,7 @@ export const FRAME_LOAD_TIMEOUT = 1000;
  * Validates that the client provided is WebdriverIO v5 or v6.
  */
 export const isWebdriverClient = (client: Browser): boolean => {
-  if (!client || typeof client !== 'object') {
+  if (!client) {
     return false;
   }
 
--- a/packages/webdriverio/src/utils.ts
+++ b/packages/webdriverio/src/utils.ts
@@ -13,7 +13,7 @@ import type {
 export const FRAME_LOAD_TIMEOUT = 1000;
 
 /**
- * Validates that the client provided is WebdriverIO v5 or v6.
+ * Validates that the client provided is WebdriverIO v5+.
  */
 export const isWebdriverClient = (client: Browser): boolean => {
   if (!client) {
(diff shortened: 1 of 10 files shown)


test/wdio/wdio-globals.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+
+describe('@wdio/globals', () => {
+  let server: Server;
+  let addr: string;
+  let builder: AxeBuilder;
+  before(async () => {
+    const app = express();
+    app.use(express.static(fixturesPath));
+    server = createServer(app);
+    addr = await testListen(server);
```
Zidious, line 17: I know we use `testListen` in the other packages for their tests, it appears it's now deprecated and moved over to [`aync-listen`](https://github.com/vercel/async-listen). We should update it here and create a tech-debt to update it in all the other packages?

test/wdio/wdio-globals.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  beforeEach(() => {
+    // node v20 causes a typescript error due to our current type of WebdriverIO.Browser not being compatible with Browser from webriverio
+    builder = new AxeBuilder({ client: browser as any });
+  });
```
Zidious, line 23: NBD - reference https://github.com/dequelabs/axe-core-npm/issues/883 so we know to update it when this gets picked up

packages/webdriverio/src/utils.ts
```diff
@@ -16,7 +16,7 @@ export const FRAME_LOAD_TIMEOUT = 1000;
  * Validates that the client provided is WebdriverIO v5 or v6.
```
Zidious, line 16: This comment is outdated we should say something like >= 5

Zidious (changes requested):
One minor comment about test-listen otherwise LGTM

test/wdio/wdio-globals.spec.ts
straker: I think that work should be done completely in a different pr.

straker: https://github.com/dequelabs/axe-core-npm/issues/884

Zidious (approved):
LGTM

PR #915, opened 2023-11-03
ci: pin GHA to specific SHA

michael-siek:


Zidious (approved):

Zidious (changes requested):
- Needs to be pinned, even though its deprecated over the new release, until we're happy to remove it, we should pin https://github.com/dequelabs/axe-core-npm/blob/develop/.github/workflows/create-release.yml#L47
-

Zidious (approved):
Ignore me 🤦  - LGTM

PR #919, opened 2023-11-04
feat: add configure method to AxeBuilder of packages/playwright

yap-house:
I want to use axe-core/playwright, but I can't seem to switch the report language ({locale}: Axe.Spec).
Therefore, I added a configure method to axe-core/playwright, which can be used the same way as axe-core/puppeteer.
However, if possible, I would like to be able to use your official packages rather than my independent modifications in actual projects.

Is it possible for you to merge this change?


dequejenn:
This is not a feature we are interested in adding at this time.

Issue #884, opened 2023-10-24
`testlisten` has been deprecated


straker:
Some of our packages use https://github.com/vercel/test-listen to get the url after starting an express server. Testlisten has been deprecated so we should move to something else. Testlisten recommends https://github.com/vercel/async-listen

PR #920, opened 2023-11-07
chore: replaced `test-listen` with `async-listen`

chrisrice202:
Closes: #884


straker (approved):
Thanks for the pr!

Approved for security (manually verified that the `async-listen` entry in the `package-lock.json` matches what it should)

PR #927, opened 2023-11-16
fix(webdriverjs): fix default commonJs export

straker:
This fixes the breaking change we accidentally released in 4.7.3 when we broke the default export of `@axe-core/webdriverjs`. Also added tests to ensure we don't make the same mistake again.

Closes https://github.com/dequelabs/axe-core-npm/issues/940

### DIFF
--- a/packages/webdriverjs/example.js
+++ b/packages/webdriverjs/example.js
@@ -1,6 +1,8 @@
-const { AxeBuilder } = require('@axe-core/webdriverjs');
+const { AxeBuilder } = require('./dist/index');
 const { Builder } = require('selenium-webdriver');
 const chrome = require('selenium-webdriver/chrome');
+
+console.log(AxeBuilder);
 
 (async () => {
   const driver = new Builder()
--- a/packages/webdriverjs/example.js
+++ b/packages/webdriverjs/example.js
@@ -1,8 +1,6 @@
-const { AxeBuilder } = require('./dist/index');
+const { AxeBuilder } = require('@axe-core/webdriverjs');
 const { Builder } = require('selenium-webdriver');
 const chrome = require('selenium-webdriver/chrome');
-
-console.log(AxeBuilder);
 
 (async () => {
   const driver = new Builder()
--- a/packages/webdriverjs/src/index.ts
+++ b/packages/webdriverjs/src/index.ts
@@ -291,4 +291,10 @@ export default class AxeBuilder {
   }
 }
 
+// ensure backwards compatibility with commonJs default export
+if (typeof module === 'object') {
+  module.exports = AxeBuilder;
+  module.exports.AxeBuilder = AxeBuilder;
+}
+
 export { AxeBuilder };
--- a/packages/webdriverjs/src/index.ts
+++ b/packages/webdriverjs/src/index.ts
@@ -294,6 +294,7 @@ export default class AxeBuilder {
 // ensure backwards compatibility with commonJs default export
 if (typeof module === 'object') {
   module.exports = AxeBuilder;
+  module.exports.default = AxeBuilder;
   module.exports.AxeBuilder = AxeBuilder;
 }
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -139,7 +139,6 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - run: npm run build --workspace=packages/react
       - run: npm run test --workspace=packages/react
       # the tests builds the project using tsc and relies on `cache.ts` to be
       # built and be it's own file. however we don't want that for the export
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
         working-directory: packages/puppeteer
       - run: npm run build --workspace=packages/puppeteer
       - run: npm run coverage --workspace=packages/puppeteer
-      - run: npm run test:esm --workspace=packages/puppeteer
+      - run: npm run test:export --workspace=packages/puppeteer
 
   cli:
     strategy:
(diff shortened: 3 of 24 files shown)


packages/webdriverjs/example.js
```diff
@@ -1,7 +1,9 @@
-const { AxeBuilder } = require('@axe-core/webdriverjs');
+const { AxeBuilder } = require('./dist/index');
 const { Builder } = require('selenium-webdriver');
 const chrome = require('selenium-webdriver/chrome');
 
+console.log(AxeBuilder);
```
stephenmathieson, line 5: 👀

stephenmathieson:
This PR touches a lot more than webdriverjs. May want to update the commit scope 🙂

packages/webdriverjs/src/index.ts
```diff
@@ -291,4 +291,10 @@ export default class AxeBuilder {
   }
 }
 
+// ensure backwards compatibility with commonJs default export
+if (typeof module === 'object') {
+  module.exports = AxeBuilder;
+  module.exports.AxeBuilder = AxeBuilder;
```
dbjorge, line 297: If someone got broken by our earlier accidental breaking change and worked around it by switching to `require('@axe-core/webdriverjs').default`, this update will break them again. I think we should avoid that.

If we instead do this...

```suggestion
  module.exports = AxeBuilder
  module.exports.default = AxeBuilder;
  module.exports.AxeBuilder = AxeBuilder;
```

...I think it allows us to support all 3 of `require()`, `require().default`, and `{ AxeBuilder } = require()` simultaneously.

I strongly think we should do that for webdriverjs, but even in all the other packages I think it would be a reasonable choice to make `require('packagename')` ergonomic.

This is the modified commonjsTest file I used to validate this approach:

```js
// ensure backwards compatibility of commonJs format
const implicitDefaultExport = require('../dist/index.js');
const explicitDefaultExport = require('../dist/index.js').default;
const { AxeBuilder } = require('../dist/index.js');
const assert = require('assert');

assert(typeof AxeBuilder === 'function', 'named export is not a function');

assert(typeof implicitDefaultExport === 'function', 'implicit default export is not a function');
assert(
  implicitDefaultExport === AxeBuilder,
  'implicit default and named export are not the same'
);

assert(typeof explicitDefaultExport === 'function', 'explicit default export is not a function');
assert(
  explicitDefaultExport === AxeBuilder,
  'explicit default and named export are not the same'
);
```

dbjorge (dismissed):
It might be nice to also include tests of importing from typescript; we've seen at least one case where a CJS/ESM-related misconfiguration resulted in a package that could be used successfully from JS, but resulted in an empty typings file that broke importing from typescript. Maybe something like:

```ts
// This file isn't executed; we only care that typescript can successfully
// build/typecheck. It is a smoke test that our build process is producing a
// basically-reasonable index.d.ts.

// Detailed tests of individual types should instead be covered by unit tests.

import { AxeBuilder as NamedImportAxeBuilder } from '../dist/index.js';
import DefaultImportAxeBuilder from '../dist/index.js';
import type { WebDriver } from 'selenium-webdriver';

// See https://stackoverflow.com/a/55541672
type IsAny<T> = 0 extends (1 & T) ? true : false;

// If the imports don't have typings assigned, these will fail
// with "ts(2322): Type 'false' is not assignable to type 'true'."
(x: IsAny<typeof NamedImportAxeBuilder>): false => x;
(x: IsAny<typeof DefaultImportAxeBuilder>): false => x;

new NamedImportAxeBuilder({} as WebDriver)
  .withRules('color-contrast')
  .analyze();

new DefaultImportAxeBuilder({} as WebDriver)
  .withRules('color-contrast')
  .analyze();
```

packages/webdriverjs/src/index.ts
```diff
@@ -291,4 +291,10 @@ export default class AxeBuilder {
   }
 }
 
+// ensure backwards compatibility with commonJs default export
+if (typeof module === 'object') {
+  module.exports = AxeBuilder;
```
dbjorge, line 296: It's unfortunate that this causes some warning spew in the build:

<details>
<summary>warning spew</summary>

```
/repos/axe-core-npm/packages/webdriverjs (webdriverjs-default-export*) » yarn build                               danbjorge@Dans-MacBook-Pro
yarn run v1.22.19
$ rimraf dist
$ tsup src/index.ts --dts --format esm,cjs
CLI Building entry: src/index.ts
CLI Using tsconfig: tsconfig.json
CLI tsup v6.7.0
CLI Target: esnext
ESM Build start
CJS Build start

 WARN  ▲ [WARNING] The CommonJS "module" variable is treated as a global variable in an ECMAScript module and may not work as expected [commonjs-variable-in-esm]

    src/index.ts:296:2:
      296 │   module.exports = AxeBuilder
          ╵   ~~~~~~

  This file is considered to be an ECMAScript module because of the "export" keyword here:

    src/index.ts:301:0:
      301 │ export { AxeBuilder };
          ╵ ~~~~~~




 WARN  ▲ [WARNING] The CommonJS "module" variable is treated as a global variable in an ECMAScript module and may not work as expected [commonjs-variable-in-esm]

    src/index.ts:296:2:
      296 │   module.exports = AxeBuilder
          ╵   ~~~~~~

  This file is considered to be an ECMAScript module because of the "export" keyword here:

    src/index.ts:301:0:
      301 │ export { AxeBuilder };
          ╵ ~~~~~~



ESM dist/index.mjs 14.24 KB
ESM ⚡️ Build success in 39ms
CJS dist/index.js 15.97 KB
CJS ⚡️ Build success in 39ms
DTS Build start
^[[A^[[ADTS ⚡️ Build success in 809ms
DTS dist/index.d.ts 2.52 KB
✨  Done in 1.27s.
```
</details>

This is a reasonable warning, but one that we reasonably want to suppress here, I think. I didn't see a good way of doing that - it comes from esbuild (not eslint), which doesn't support a suppress-next-line type comment. esbuild does have a [`--log-override`](https://esbuild.github.io/api/#log-override) option to quiet it, but only for the whole build process, not for a few lines, and tsup doesn't support forwarding it anyway ([they tried](https://github.com/egoist/tsup/pull/932) but encountered some issues getting it to work and abandoned it :( )

I don't think it's worth more time trying to suppress the warning, just noting it here so noone repeats the same research.

straker:
I'm going to split the TS type check and the support for default/`.default`/named export into their own prs, though I will add the default/`.default`/named export into webdriverjs to ensure we don't break 4.7.3 with 4.7.4.

PR #943, opened 2023-11-21
chore: allow csj to export a default value

straker:
This adds an esbuild plugin to "fix" the esbuild problem of having `export default a` changed into `module.exports.default = `` instead of `modules.exports = a`. Using the plugin removes the build warnings we get from esbuild when we tried to add `module.exports` into the code to do this manually.

### DIFF
--- a/utils/esbuild-plugin-cjs-interop.ts
+++ b/utils/esbuild-plugin-cjs-interop.ts
@@ -1,0 +1,54 @@
+import path from 'path';
+
+/**
+ * "Fixes" the esbuild problem of exporting the CJS default export as `module.exports.default`
+ * instead of `module.exports`. The plugin appends a block of code to the file that takes the
+ * `.default` module export and re-exports it as `module.exports`. It then takes all named
+ * exports and re-exports them as part of the `module.exports` under the same name. This also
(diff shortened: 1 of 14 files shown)


utils/esbuild-plugin-cjs-interop.ts
```diff
@@ -0,0 +1,54 @@
+import path from 'path';
+
+/**
+ * "Fixes" the esbuild problem of exporting the CJS default export as `module.exports.default`
+ * instead of `module.exports`. The plugin appends a block of code to the file that takes the
+ * `.default` module export and re-exports it as `module.exports`. It then takes all named
+ * exports and re-exports them as part of the `module.exports` under the same name. This also
+ * gives the benefit of exporting the `.default` module which allows us to support all 3 export
+ * styles: the default export, `.default` export, and named exports.
```
WilcoFiers, line 9: Is there an issue with ESBuild open on this? If not, could you submit one? This feels like it would be a generally useful feature.

WilcoFiers (dismissed):
Small nitpick, up to you.

WilcoFiers (approved):

Issue #829, opened 2023-09-18
webdriverio: Update to latest webdriverio


straker:
As part of [the conversion to npm workspaces](https://github.com/dequelabs/axe-core-npm/pull/687) we had to [pin the version of Webdriverio](https://github.com/dequelabs/axe-core-npm/pull/687#discussion_r1322081470) as the latest version causes type errors. Need to spend time updating the types.

dbjorge:
Notably, this will including updating the package to rely on the `Browser`/`Element` types from the global `WebdriverIO.*` type namespace, rather than the ones exported from the index of the `webdriverio` package - the latter was deprecated in https://github.com/webdriverio/webdriverio/commit/2a66834912741fbc583e7f544571ce1d6cb0dfcc and is expected to be removed in webdriverio v9.

We'll need to test how to make that cooperate with back-compat for old webdriverio versions we support - I suspect it might involve type shenanigans like `Awaited<ReturnType<webdriverio.remote>> | WebdriverIO.Browser`.

2erix:
We are now impacted in our organisation because the AxeBuilder requires deprecated interface `Browser` type from npm package/path `WebdriverIO/build/types` which we use as browser/driver in our projects. 

WDIO now supplies browser/driver types from  package/path `@wdio/globals/types` as `WebdriverIO.Browser` instead. 

Currently we don’t have any solution except wait for an update to AxeBuilder, so anything that could be expedited would be helpful. 

In the meantime I am using the deprecated type at risk.

padmavemulapati:
Verified with the latest Canary version - `4.8.3-6b60442.0` and with typescript- webdriverio is working as expected. Validated with V8, V7 and  V6 . With all the versions typescript is working fine.

Steps validated:
1. installed @axe-core/webdriverio package in webdriverio (5.3.3 , v6.12.1, v7.34.0, v8.21.0, and @wdio/cli- v8.24.2,
    @wdio/globals-8.24.2 ) 
 

Test code: compile and run using typescript 
```

import { AxeBuilder } from '@axe-core/webdriverio'
import { remote } from 'webdriverio'

(async () => {
  const client = await remote({
    logLevel: 'error',
    capabilities: {
      browserName: 'chrome',
      'goog:chromeOptions': {
        args: ['headless', 'disable-gpu']
      }
    }
  })
  await client.url('https://dequeuniversity.com/demo/mars/')

  try {
    const results = await new AxeBuilder({ client }).analyze()
[...]

PR #962, opened 2023-12-20
fix(webdriverio): fix types between v8 and <v8

straker:
This change brings the types from webdriverio v8 and makes them compatible with the types from <= v7. This change also prepares us to take the latest webdriverio v8 type changes which were causing our builds to break.

This also upgrades the version of `@wdio` packages in `test/wdio` in order to not have multiple copies across packages. We should update all these types together so they don't go out-of-sync.

**QA notes:**

With this change we should be able to run `@axe-core/webdriverio` package in webdriverio v5+. Using the following code, install each version of webdriverio from v5 to v8, and then compile and run using typescript to verify working

```ts
import { AxeBuilder } from '@axe-core/webdriverio'
import { remote } from 'webdriverio'

(async () => {
  const client = await remote({
    logLevel: 'error',
    capabilities: {
      browserName: 'chrome',
      'goog:chromeOptions': {
        args: ['headless', 'disable-gpu']
      }
    }
  })
  await client.url('https://dequeuniversity.com/demo/mars/')

  try {
    const results = await new AxeBuilder({ client }).analyze()
    console.log(results)
  } catch (e) {
    console.error(e)
  }

  client.deleteSession()
})()
```

**Additional notes:**

Webdriverio v7 cannot run in node v18/20, so you'll need to use node v16 in order to test <=v7. For webdriverio v5, the code above should compile but may not open a browser and run. This is a set up issue with the webdriverio and not our code so we are OK if the builds but doesn't run in v5.

Closes: https://github.com/dequelabs/axe-core-npm/issues/829


.github/workflows/tests.yml
```diff
@@ -59,7 +59,7 @@ jobs:
       - run: npm ci
         # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
       - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chromedriver --verbose
+        run: npx browser-driver-manager install chromedriver=119 --verbose
```
straker, line 62: This is a temporary fix that will be fix in a separate pr to fix the cli tests with 120 https://github.com/dequelabs/axe-core-npm/issues/963.

packages/webdriverio/src/index.ts
```diff
@@ -246,15 +253,15 @@ export default class AxeBuilder {
     // ensure we fail quickly if an iframe cannot be loaded (instead of waiting
     // the default length of 30 seconds)
     const { pageLoad } = await this.client.getTimeouts();
-    this.client.setTimeout({
+    (this.client as WebdriverIO.Browser).setTimeout({
```
straker, line 256: This fixes a problem with older versions of webdriverio as in newer versions they declare the function with the first parameter as [this: Webdriverio.Browser](https://github.com/webdriverio/webdriverio/blob/main/packages/webdriverio/src/commands/browser/setTimeout.ts#L40).

packages/webdriverio/src/utils.ts
```diff
@@ -130,13 +130,13 @@ async function assertFrameReady(client: Browser): Promise<void> {
...
 
 export const axeRunPartial = (
-  client: Browser,
+  client: WdioBrowser,
   context?: SerialContextObject,
   options?: RunOptions
 ): Promise<PartialResult> => {
   return promisify(
-    client
-      .executeAsync<string, []>(
```
straker, line 139: Older versions of webdriverio [don't declare the type](https://github.com/webdriverio/webdriverio/blob/v6.12.1/packages/webdriverio/src/commands/browser/executeAsync.ts#L46) of the return so we can't declare it either. This defaults to `<unknown, []>` for the type in newer versions which is why we need to make the change below to make the string type `unknown` and then cast it to a `string`

packages/webdriverio/src/utils.ts
```diff
@@ -187,12 +187,12 @@ export const axeRunLegacy = (
         callback(JSON.stringify(axeResults))
       });`
       )
-      .then((r: string) => deserialize<AxeResults>(r))
+      .then((r: unknown) => deserialize<AxeResults>(r as string))
```
Zidious, line 190: nit - this doesn't need to be cast as `unknown` it's already that type

Zidious (changes requested):
Looking good, just a few minor comments inline -

packages/webdriverio/src/types.ts
```diff
@@ -1,14 +1,39 @@
...
+      | 'getWindowHandle'
+      | 'switchToWindow'
+      | 'createWindow'
+      | 'url'
+      | 'getTimeouts'
+      | 'setTimeout'
+      | 'closeWindow'
+      | 'executeAsync'
+      | 'execute'
+    >;
```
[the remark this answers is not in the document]
straker: No. It needs to be picked as there are incompatible types between the two. Luckily they are incompatible in functions we don't use so we can just pick them out.

Zidious (approved):
LGTM!

PR #1000, opened 2024-01-29
added access to spec option for wdio

udarrr:
There is no possibility to have access to 'noHTML' option that's why all outcomes printing to console, it's not suitable in ci environment, having access prevent the behavior.


udarrr:
@michael-siek  @straker could you review it ?

straker:
This for the pr. Unfortunately this is not a feature we are interested in adding at this time.

udarrr:
@straker I'm a bit confused what's wrong with it ?

PR #1019, opened 2024-02-13
ci: generate notes from tag

michael-siek:
This should generate the notes based on tag

no qa required


.github/workflows/deploy.yml
```diff
@@ -86,7 +86,7 @@ jobs:
           PKG_VERSION=$(jq -r .version < lerna.json)
           gh release create v$PKG_VERSION \
             --title "Release $PKG_VERSION" \
-            --generate-notes \
+            --notes-file CHANGELOG.md \
```
Zidious, line 89: Does this append the _entire_ changelog or only new entries?

Zidious (changes requested):
One comment inline -

Zidious (approved):

Issue #936, opened 2023-11-20
WebdriverJS with the Safari driver causes noSuchWindowException


Zidious:
### Product

webdriverjs

### Product Version

4.8.0

### Latest Version

- [X] I have tested the issue with the latest version of the product

#### Expectation
Print axe results 

#### Actual
Throws an error 

#### How to Reproduce

```js
const { AxeBuilder } = require("@axe-core/webdriverjs");
const { Builder } = require("selenium-webdriver");

(async () => {
  const driver = new Builder().forBrowser("safari").build();
  await driver.get("https://www.google.com");

  const results = await new AxeBuilder(driver).analyze();
  console.log(results);

  await driver.quit();
})();
```

#### Additional context
Run into the same issue in [Ruby](https://github.com/dequelabs/axe-core-gems/issues/352) and [Java](https://github.com/dequelabs/axe-core-maven-html/pull/392). 

We should look to to removing: https://github.com/dequelabs/axe-core-npm/blob/5fcb30f8be6bdcd196484384efd0e036b0bc5184/packages/webdriverjs/src/index.ts#L276-L285

In favour of something like: 

```ts
try {
  await driver.switchTo().newWindow('tab')
  await driver.get('about:blank');
} catch (error) {
  throw new Error(
    `switchTo failed. Are you using updated browser drivers? \nDriver reported:\n${error}`
  );
}
```

Selenium doesn't appear to document the above in their [docs](https://www.selenium.dev/selenium/docs/api/javascript/module/selenium-webdriver/lib/webdriver_exports_TargetLocator.html), but it's available within their [source code](https://github.com/SeleniumHQ/selenium/blob/eb0a32195277dfb6f058ce7ed6cd16322bd36067/javascript/node/selenium-webdriver/lib/webdriver.js#L2484)

padmavemulapati:
Verified this with the latest RC-build (4.9.1-4cb64b0.0 )
I am seeing below error when executing on safari (working fine on chrome and firefox)
```
dq_padmavathi@DQPadmavathi-MacBook-Pro axe-core-npm % node tests/wdjs.js               
/Users/dq_padmavathi/Desktop/DevtoolsAutomation/qa-axe-devtools-api-daf/axe-core-npm/node_modules/selenium-webdriver/remote/index.js:256
              let cancelToken = earlyTermination.catch((e) => reject(Error(e.message)))
                                                                     ^

Error: Server terminated early with status 1
[...]

PR #1027, opened 2024-03-22
fix: WebdriverJS with Safari can cause noSuchWindowException

scottmries:
Reproing according to the associated issue would occasionally throw a `noSuchWindowException`, about one in three times; this change should fix that.

Closes: #936


Zidious (changes requested):
We'd want to scope this fix to webdriverjs so it'd only appear in that package changelog i.e. `fix(webdriverjs): ...` 

Left a comment inline

PR #1031, opened 2024-03-28
fix: get correct new window handle with Selenium 3 workaround

scottmries:
This is the workaround fix that supports Selenium 3 and 4. We would likely switch to #1027 should we drop Selenium 3 support. 
Rather than always using the last handle after opening a new window, we filter for the unique new handle and use it. This is analogous to the [axe-core-maven-html approach](https://github.com/dequelabs/axe-core-maven-html/blob/ad58b26a8d0e2f1afed33b2c5cbca22b54644b99/selenium/src/main/java/com/deque/html/axecore/extensions/WebDriverExtensions.java#L126).

Closes: #936


packages/webdriverio/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/express": "^4.17.14",
     "@types/mocha": "^10.0.6",
     "@types/node": "^20.8.10",
-    "@wdio/globals": "^8.27.0",
+    "@wdio/globals": "^8.32.2",
```
michael-siek, line 61: I'm not sure this is relevant for this PR

packages/webdriverjs/test/axe-webdriverjs.spec.ts
```diff
@@ -1,16 +1,17 @@
 import 'mocha';
 import type { AxeResults, Result } from 'axe-core';
-import type { WebDriver } from 'selenium-webdriver';
+import { WebDriver } from 'selenium-webdriver';
```
michael-siek, line 3:
```suggestion
import type { WebDriver } from 'selenium-webdriver';
```

michael-siek (changes requested):
Small nitpick

packages/webdriverjs/test/axe-webdriverjs.spec.ts
michael-siek: It looks like `WebDriver` is only still being used as a type

michael-siek (approved):
LGTM

PR #1038, opened 2024-04-02
test(cli): add test to cover selenium-webdriver <4.17.0

michael-siek:
Add a test to cover the event that users are using selenium-webdriver <4.17.0 and chrome options use the "old" way of supporting headless runs. Also did some cleanup and removed code that wasn't required in cli

no qa required

### DIFF
--- a/packages/cli/src/lib/axe-test-urls.ts
+++ b/packages/cli/src/lib/axe-test-urls.ts
@@ -19,7 +19,7 @@ const testPages = async (
     const currentUrl = urls[0].replace(/[,;]$/, '');
 
     if (events?.onTestStart) {
-      events?.onTestStart(currentUrl);
+      events.onTestStart(currentUrl);
     }
 
     if (config.timer) {
--- a/packages/cli/src/lib/utils.test.ts
+++ b/packages/cli/src/lib/utils.test.ts
@@ -3,7 +3,6 @@ import { assert } from 'chai';
 import tempy from 'tempy';
 import { join } from 'path';
 import { mkdirSync, writeFileSync, rmSync } from 'fs';
-import { dependencies } from '../../package.json';
 import * as utils from './utils';
 
 describe('utils', () => {
@@ -83,11 +82,7 @@ describe('utils', () => {
         writeFileSync(join(parentDirname, 'axe.js'), 'parent');
 
         const cliDirname = join(tempDir, 'packages', 'cli');
-        const nodeModDirname = join(
-          cliDirname,
-          'node_modules',
-          'axe-core'
-        );
+        const nodeModDirname = join(cliDirname, 'node_modules', 'axe-core');
         mkdirSync(nodeModDirname, { recursive: true });
         writeFileSync(join(nodeModDirname, 'axe.js'), 'node modules');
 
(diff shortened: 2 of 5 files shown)


packages/cli/src/lib/axe-test-urls.ts
```diff
@@ -19,7 +19,7 @@ const testPages = async (
     const currentUrl = urls[0].replace(/[,;]$/, '');
 
     if (events?.onTestStart) {
-      events?.onTestStart(currentUrl);
+      events.onTestStart(currentUrl);
```
michael-siek, line 22: optional chaining is not required

packages/cli/src/lib/utils.test.ts
```diff
@@ -3,7 +3,6 @@ import { assert } from 'chai';
 import tempy from 'tempy';
 import { join } from 'path';
 import { mkdirSync, writeFileSync, rmSync } from 'fs';
-import { dependencies } from '../../package.json';
```
michael-siek, line 6: This wasn't used

packages/cli/src/lib/webdriver.test.ts
```diff
@@ -4,9 +4,10 @@ import { startDriver } from './webdriver';
 import { WebDriver } from 'selenium-webdriver';
 import chromedriver from 'chromedriver';
 import chrome from 'selenium-webdriver/chrome';
-import type { Options } from 'selenium-webdriver/chrome';
```
michael-siek, line 7: Also wasn't used

packages/cli/src/lib/webdriver.test.ts
```diff
@@ -22,7 +23,9 @@ describe('startDriver', () => {
   });
 
   afterEach(async () => {
-    await driver.quit();
+    try {
+      await driver.quit();
+    } catch (error) {}
```
Zidious, line 28: Is the try-catch required here?

packages/cli/src/lib/webdriver.test.ts
```diff
@@ -101,4 +104,15 @@ describe('startDriver', () => {
...
     assert.deepEqual(timeoutValue.script, 10000000);
   });
+
+  it('test headless option in selenium-webdriver < 4.17.0', async () => {
+    const stub = sinon.stub(chrome, 'Options').returns({
+      headless: () => {}
+    });
+    try {
+      driver = await startDriver(config);
+    } catch (error) {}
```
Zidious, line 114: Is the try-catch required here?

packages/cli/src/lib/webdriver.test.ts
```diff
@@ -101,4 +104,15 @@ describe('startDriver', () => {
     assert.isObject(timeoutValue);
     assert.deepEqual(timeoutValue.script, 10000000);
   });
+
+  it('test headless option in selenium-webdriver < 4.17.0', async () => {
```
Zidious, line 108: nit - we aren't testing if the `options.headless()` call passes the right flag to the Driver, we are just asserting that the function was called when using < 4.17.0 - 
```suggestion
  it('invokes `options.headless()` on versions of selenium-webdriver < 4.17.0', async () => {
```

Zidious (changes requested):
Minor things below -

packages/cli/src/lib/webdriver.test.ts
michael-siek: Ya it looks like this is required since driver isn't built properly due to mocked chrome options

Zidious: So we don't forget can we add comments to why it was needed and for [L114](https://github.com/dequelabs/axe-core-npm/pull/1038/files#diff-48a6ca6c15e8bb1212533365531f7a5e95434f92acf588c23e62fb2c64f7fdaeR114)

Zidious (approved):

PR #1079, opened 2024-06-04
chore: fix eslint errors and warnings

straker:
Fixes all the lint errors and warnings in preparation to #1078

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -33,7 +33,8 @@ module.exports = [
       '@typescript-eslint/ban-types': 'off',
       '@typescript-eslint/no-unused-vars': 'error',
       '@typescript-eslint/no-explicit-any': 'error',
-      '@typescript-eslint/no-non-null-assertion': 'error'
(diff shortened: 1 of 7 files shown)


eslint.config.js
```diff
@@ -33,7 +33,8 @@ module.exports = [
       '@typescript-eslint/ban-types': 'off',
       '@typescript-eslint/no-unused-vars': 'error',
       '@typescript-eslint/no-explicit-any': 'error',
-      '@typescript-eslint/no-non-null-assertion': 'error'
+      '@typescript-eslint/no-non-null-assertion': 'error',
+      '@typescript-eslint/no-require-imports': 'off'
```
straker, line 37: We use imports or requires based on the file (test files, examples, eslint config itself, postinstall, etc.). Instead of having to disable it or ignore it for every file we want, easier to just disable the rule and let us decide when to use require or import per file.

Zidious (approved):

Issue #1043, opened 2024-04-08
Update how repo downloads Chromedriver


dequejenn:
### Upgrade version of `browser-driver-manager` in each repository CI configuration

`browser-driver-manager` should upgrade its version so that the repos can use the new chrome driver installation method.

## Update repos that downloads `chromedriver`

Each repo that downloads `chromedriver` will need to do it using `browser-driver-manager`

Please reference [Usage guide](https://www.npmjs.com/package/browser-driver-manager#usage)

## Usage

When using this chromedriver path in the tests, you can use the following code to get the path of the chromedriver:

## Getting the path of chromedriver

To be able to get the path of where chromedriver exist in non-node libraries each library can read the `.env` file created by `browser-driver-manager` and get the path from there.

### NPM based packages

```javascript
const { config } = require('dotenv')
const os = require('os')
const path = require('path')

const HOME_DIR = os.homedir()
const BDM_CACHE_DIR = path.resolve(HOME_DIR, '.browser-driver-manager')

config({ path: path.resolve(BDM_CACHE_DIR, '.env') })

const { CHROMEDRIVER_TEST_PATH, CHROME_TEST_PATH } = process.env

```

### Non-NPM based packages (Psuedo code)

```python
from os import path
from dotenv import dotenv_values

env = dotenv_values(path.join(path.expanduser('~'), '.browser-driver-manager', '.env'))
chromedriver_path = env['CHROMEDRIVER_TEST_PATH']
chrome_path = env['CHROME_TEST_PATH']
```

```ruby
require 'dotenv'

Dotenv.load(File.join(Dir.home, '.browser-driver-manager', '.env'))
chromedriver_path = ENV['CHROMEDRIVER_TEST_PATH']
chrome_path = ENV['CHROME_TEST_PATH']
```

```java
import java.util.Map;
import java.util.HashMap;
import java.io.File;
import java.io.FileReader;
import java.io.BufferedReader;
import java.io.IOException;

public class GetChromePath {
  public static void main(String[] args) {
    Map<String, String> env = new HashMap<String, String>();
    try {
      BufferedReader reader = new BufferedReader(new FileReader(new File(System.getProperty("user.home") + "/.browser-driver-manager/.env")));
      String line;
      while ((line = reader.readLine()) != null) {
        String[] parts = line.split("=");
        env.put(parts[0], parts[1]);
      }
      reader.close();
    } catch (IOException e) {
      e.printStackTrace();
    }
[...]

PR #1080, opened 2024-06-07
chore: use updated browser-driver-manager

scottmries:
No QA required

Closes: #1043

### DIFF
--- a/packages/puppeteer/test/utils.ts
+++ b/packages/puppeteer/test/utils.ts
@@ -4,6 +4,7 @@ import listen from 'async-listen';
 import { expect } from 'chai';
 import type { PuppeteerLaunchOptions } from 'puppeteer';
 import { fixturesPath } from 'axe-test-fixtures';
+import chromedriver from 'chromedriver';
 
 export async function expectAsync(
   fn: () => Promise<any>
@@ -42,7 +43,7 @@ export function puppeteerOpts(): PuppeteerLaunchOptions {
   if (process.env.CI) {
     options.args = [];
     options.args.push('--no-sandbox', '--disable-setuid-sandbox');
-    options.executablePath = '/usr/bin/google-chrome-stable';
+    options.executablePath = chromedriver.path;
   }
 
   return options;
--- a/packages/puppeteer/test/utils.ts
+++ b/packages/puppeteer/test/utils.ts
@@ -4,7 +4,6 @@ import listen from 'async-listen';
 import { expect } from 'chai';
 import type { PuppeteerLaunchOptions } from 'puppeteer';
 import { fixturesPath } from 'axe-test-fixtures';
-import chromedriver from 'chromedriver';
 
 export async function expectAsync(
   fn: () => Promise<any>
@@ -43,7 +42,6 @@ export function puppeteerOpts(): PuppeteerLaunchOptions {
   if (process.env.CI) {
     options.args = [];
     options.args.push('--no-sandbox', '--disable-setuid-sandbox');
-    options.executablePath = chromedriver.path;
   }
 
   return options;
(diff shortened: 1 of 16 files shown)


packages/webdriverio/test/axe-webdriverio.spec.ts
```diff
@@ -53,8 +59,10 @@ describe('@axe-core/webdriverio', () => {
       let chromedriverProcess: ChildProcessWithoutNullStreams;
 
       before(async () => {
-        const path = process.env.CHROMEDRIVER_PATH ?? chromedriver.path;
-        chromedriverProcess = child_process.spawn(path, [`--port=${port}`]);
+        const path = process.env.CHROMEDRIVER_TEST_PATH;
```
dbjorge, line 62: Shouldn't this also be specifying `CHROME_TEST_PATH` somewhere?

packages/webdriverjs/test/test-utils.ts
```diff
@@ -13,9 +20,7 @@ export const Webdriver = (): WebDriver => {
     .forBrowser('chrome')
     .setChromeOptions(options)
     .setChromeService(
-      new chrome.ServiceBuilder(
-        process.env.CHROMEDRIVER_PATH || chromedriver.path
-      )
+      new chrome.ServiceBuilder(process.env.CHROMEDRIVER_TEST_PATH)
```
dbjorge, line 23: Shouldn't this also be specifying `CHROME_TEST_PATH` somewhere?

packages/puppeteer/test/utils.ts
```diff
@@ -42,7 +49,7 @@ export function puppeteerOpts(): PuppeteerLaunchOptions {
   if (process.env.CI) {
     options.args = [];
     options.args.push('--no-sandbox', '--disable-setuid-sandbox');
-    options.executablePath = '/usr/bin/google-chrome-stable';
+    options.executablePath = process.env.CHROME_TEST_PATH;
```
dbjorge, line 52: I think we probably want cases where the browser automation framework bundles a synced version of the browser (playwright, puppeteer) to use that synced version, so I don't think this should actually be going through BDM (and I don't think the original should have been going through the `/usr/bin/` version, either)

dbjorge (changes requested):

packages/webdriverio/test/axe-webdriverio.spec.ts
```diff
@@ -110,7 +118,7 @@ describe('@axe-core/webdriverio', () => {
             browserName: 'chrome',
             'goog:chromeOptions': {
               args: ['--headless'],
-              binary: binaryPath
+              binary: binaryPath ?? (process.env.CHROME_TEST_PATH as string)
```
scottmries, line 121: I'm not entirely clear on the need for the Windows path here, but I'm deferring to it just in case.

dbjorge: I think this should probably be dropped; it should be unnecessary now that `CHROME_TEST_PATH` is controlling the browser+driver version. Even on Windows, we want the tests to use the BDM-synced version of chrome that matches the version of chromedriver, so I don't think we should allow windows to use the globally-installed chrome version.

dbjorge (changes requested):

packages/webdriverjs/test/test-utils.ts
```diff
@@ -1,21 +1,27 @@
...
+const BDM_CACHE_DIR = path.resolve(HOME_DIR, '.browser-driver-manager');
+
+config({ path: path.resolve(BDM_CACHE_DIR, '.env') });
 
 export const Webdriver = (): WebDriver => {
   // Weird type change since 4.23.1 release
   // @see https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/69724
   const options = new chrome.Options();
   options.addArguments('headless');
+  options.setBinaryPath(process.env.CHROME_TEST_PATH as string);
```
dbjorge, line 18: What happens if the user doesn't run BDM before running these tests? I think right now it'll fall back to the globally installed versions, but I think it would be preferable to bail early with an actionable error message that explains the required BDM command; I think we'd prefer to force local devs to use the same setup we'll be testing against in CI.

dbjorge (changes requested):

dbjorge (approved):

PR #1098, opened 2024-08-02
chore: downgrade lerna to resolve ESM requirement

michael-siek:
no qa required

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "eslint-plugin-prettier": "^5.1.3",
         "globals": "^15.2.0",
         "husky": "^9.0.10",
-        "lerna": "^8.1.3",
+        "lerna": "8.1.3",
         "lint-staged": "^15.1.0",
         "prettier": "^3.0.3",
         "react": "^17.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "eslint-plugin-prettier": "^5.1.3",
         "globals": "^15.2.0",
         "husky": "^9.0.10",
-        "lerna": "^8.0.1",
+        "lerna": "^8.1.3",
         "lint-staged": "^15.1.0",
         "prettier": "^3.0.3",
         "react": "^17.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "eslint-plugin-prettier": "^5.1.3",
     "globals": "^15.2.0",
     "husky": "^9.0.10",
-    "lerna": "^8.0.1",
+    "lerna": "^8.1.3",
```
Zidious, line 25: Was this intended? We're not downgrading here 👀

Zidious (changes requested):

package-lock.json
```diff
@@ -3397,76 +3419,71 @@
       }
     },
     "node_modules/@lerna/create": {
-      "version": "8.1.5",
-      "resolved": "https://registry.npmjs.org/@lerna/create/-/create-8.1.5.tgz",
-      "integrity": "sha512-Ku8yTGgeumayvMr8sml72EPb6WaoJhRjMTkMZrKSJtcLNDBlDpKwyUxDxNTBNBRUYWUuJCnj7eUH7pDNuc9odQ==",
+      "version": "8.1.3",
```
michael-siek, line 3422: @Zidious sorry the downgrade existed in the lock file.

Zidious: Ah missed that! Should be pin Lerna then to 8.1.3? Dependabot will update this next time around to >8.1.5

michael-siek: Good call just updated and pinned to 8.1.3

package.json
```diff
@@ -22,7 +22,7 @@
     "eslint-plugin-prettier": "^5.1.3",
     "globals": "^15.2.0",
     "husky": "^9.0.10",
-    "lerna": "^8.0.1",
+    "lerna": "8.1.3",
```
Zidious, line 25: Will need to add it to the `dependabot.yml` too

Zidious (dismissed):

Zidious (approved):

PR #1150, opened 2025-01-28
ci: resolve chromedrivers not running in Ubuntu runners >=24.04

Zidious:
This patch adds `--no-sandbox` chrome flag which was already added by default on previous versions of bun

Resolves this [error](https://github.com/dequelabs/axe-core-npm/actions/runs/12889457050/job/36296592750#step:6:403): 

> SessionNotCreatedError: session not created: probably user data directory is already in use, please specify a unique value for --user-data-dir argument, or don't use --user-data-dir

Related issues: https://github.com/SeleniumHQ/selenium/issues/14609

No QA required


scottmries (approved):
Excellent sleuthing. LGTM 🎉

PR #1158, opened 2025-03-18
chore: update node to 22

scottmries:
Closes https://github.com/dequelabs/bass-team/issues/176

No QA required


scottmries:
It looks like the tests for Node 16 still want to run, but they no longer exist in the`tests.yml`.

stephenmathieson (approved):
LGTM

We'll need to update the branch requirements before merging, as we explicitly require the "16" versions of the tests to pass. There is probably a better way to do this moving forward.

image

Issue #1130, opened 2024-10-25
webdriverio peer dep needs to be updated to include v9


smackfu:
### Product

webdriverio

### Product Version

4.10.0

### Latest Version

- [x] I have tested the issue with the latest version of the product

#### Expectation
If I have the latest version of webdriverio installed, I would expect @axe-core/webdriverio to use it and not get a peer dep error.

#### Actual
Peer dep list only include v8 so there is a peer dep error.
```
$ npm install --save-dev @axe-core/webdriverio@latest
npm error code ERESOLVE
npm error ERESOLVE unable to resolve dependency tree
npm error
npm error While resolving: example@1.0.0
npm error Found: webdriverio@9.2.1
npm error node_modules/webdriverio
npm error   dev webdriverio@"^9.2.1" from the root project
npm error
npm error Could not resolve dependency:
npm error peer webdriverio@"^5 || ^6 || ^7 || ^8" from @axe-core/webdriverio@4.10.0
npm error node_modules/@axe-core/webdriverio
npm error   dev @axe-core/webdriverio@"4.10.0" from the root project
npm error
npm error Fix the upstream dependency conflict, or retry
npm error this command with --force or --legacy-peer-deps
npm error to accept an incorrect (and potentially broken) dependency resolution.
```

#### How to Reproduce

1. `npm init`
2. `npm install --save-dev webdriverio@latest`
3. `npm install --save-dev @axe-core/webdriverio@latest`

#### Additional context

webdriverio was bumped to v9 on Aug 15, and has had several releases so should be stable by now:
https://www.npmjs.com/package/webdriverio/v/9.0.0

padmavemulapati:
Able to run the CoreNPM-supported APIs working with WebdriverIO v9 


Uploading Screen Recording 2025-05-15 at 7.02.57 PM.mov…

we need to add browser capabilities like below:

```
capabilities: {
        browserName: 'chrome',
        'wdio:enforceWebDriverClassic': true,
        'goog:chromeOptions': {
          args: []
        }
      }
```

smackfu:
This was duplicated by https://github.com/dequelabs/axe-core-npm/issues/1160 and fixed by https://github.com/dequelabs/axe-core-npm/pull/1167

PR #1169, opened 2025-04-24
fix(wdio): resolve blank navigation issue in WDIO v9

scottmries:
No QA required.


padmavemulapati:
Added comment here - https://github.com/dequelabs/axe-core-npm/issues/1130

PR #683, opened 2023-03-11
fix(cli): increase --timeout default to 90 seconds (to match --help)

not-my-profile:
The timeout was decreased in 6f5498ee8a2acb78846aa623773c7d1de19125e0 to 20 seconds (apparently accidentally since that commit was classfied as a refactor ... and it also didn't update the --help text accordingly).

This commit changes it back to 90 seconds to match the --help text.

PR #1207, opened 2025-08-11
chore: fix typescript issues

Garbee:
Fix a few ts-ignore issues throughout the system. One of these is biting us right now in CI. The others are just candidates for future issues we don't need to keep around.

No QA required

### DIFF
--- a/packages/webdriverio/test/axe-webdriverio.spec.ts
+++ b/packages/webdriverio/test/axe-webdriverio.spec.ts
@@ -15,6 +15,7 @@ import { ChildProcessWithoutNullStreams } from 'child_process';
 import { fixturesPath } from 'axe-test-fixtures';
 import { config } from 'dotenv';
 import os from 'os';
+import sinon from 'sinon';
 
 const HOME_DIR = os.homedir();
 const BDM_CACHE_DIR = path.resolve(HOME_DIR, '.browser-driver-manager');
@@ -139,6 +140,7 @@ describe('@axe-core/webdriverio', () => {
       });
 
       afterEach(async () => {
+        sinon.restore();
         await client.deleteSession();
         server.close();
       });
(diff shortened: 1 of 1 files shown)


packages/webdriverio/test/axe-webdriverio.spec.ts
```diff
@@ -1411,13 +1411,8 @@ describe('@axe-core/webdriverio', () => {
...
 
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
-          delete client.createWindow;
-          // eslint-disable-next-line @typescript-eslint/ban-ts-comment
-          // @ts-ignore
-          client.createWindow = () => {
-            return { handle: null };
+          client.createWindow = (type: 'tab' | 'window') => {
+            return Promise.resolve({ handle: '', type });
```
Garbee, line 1415: Just replacing the method with a properly typed stub to match the underlying signature.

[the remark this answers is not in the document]
Garbee: The underlying type signature is:

```typscript
 /**
     * Webdriver Protocol Command
     *
     * Create a new top-level browsing context.
     * @ref https://w3c.github.io/webdriver/#new-window
     *
     */
    createWindow(type: 'tab' | 'window'): Promise<WindowHandle>;
```

This is correct.

packages/cli/src/bin/cli.ts
```diff
@@ -58,12 +58,7 @@ program
     '--load-delay <n>',
     'Set how much time (milliseconds) axe will wait after page load before running the audit (default: 0)'
   )
-  .option(
-    '--timeout <n>',
-    'Set how much time (seconds) axe has to run',
-    // @ts-ignore
-    90
-  )
+  .option('--timeout <n>', 'Set how much time (seconds) axe has to run', '90')
```
zlayaAvocado, line 61: Do you have any idea why it was passed as a number previously? Seems like an intentional thing to me for whatever reason, especially since it was ts-ignored 🤔

Garbee: The [original pr](https://github.com/dequelabs/axe-core-npm/pull/683) has no useful context on that. Looks to me like they just knew it should be a number so they made it a number by force. When if you use `--timeout 120` in the terminal, it is still coming in as a string and needs to be accounted for. The math is casting correctly, so it works as a string.

PR #1242, opened 2025-10-27
chore: fix build script

### LINKED ISSUE #1238: Update deployment workflow to use OIDC
The publish workflow needs to be updated to utilize OIDC publishing. It is already all consolidated into one deployment file, so the update should be fairly minimal.

Garbee:
Only run build scripts if present in the workspaces. For the test workspaces that don't need a build, this removes those that are defined as just empty echos.

Fixes: #1238

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "lint": "eslint .",
     "prepare": "husky install",
     "precommit": "lint-staged",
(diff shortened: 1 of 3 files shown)


scottmries (approved):
TIL about `--if-present`. Nice work.

PR #1261, opened 2025-12-21
fix: reorder exports to place types first

### LINKED ISSUE #1243: Fix export list order
The packages seem to have the exports defined in an improper order. The types for example are last, which means they won't get imported by a consumer.

## Acceptance Criteria

All packages align to the [conditional exports](https://nodejs.org/api/packages.html#conditional-exports) recommended order from Node.

## Useful information

The packages give warnings through [tsup output](https://github.com/dequelabs/axe-core-npm/actions/runs/18844944995/job/53766410399#step:6:30) during the build process. In the end of this, the build should be free of warnings.

JustasMonkev:
Move the "types" condition to the first position in package.json
  exports field to align with Node.js conditional exports standards.
  This ensures TypeScript properly resolves type definitions.

Closes:  #1243

### DIFF
--- a/packages/playwright/package.json
+++ b/packages/playwright/package.json
@@ -31,9 +31,9 @@
   "types": "./dist/index.d.ts",
   "exports": {
     ".": {
+      "types": "./dist/index.d.ts",
       "import": "./dist/index.mjs",
-      "require": "./dist/index.js",
-      "types": "./dist/index.d.ts"
+      "require": "./dist/index.js"
     }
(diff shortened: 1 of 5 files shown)


straker (approved):
Thanks for the changes.

Garbee (approved):
The ordering is correct per Node docs.

Thank you for putting the time into this.

straker:
Approved for security.

PR #1266, opened 2025-12-23
chore: rebuild lockfile

Garbee:
After a few dependabot updates, the canary deploy started failing. This updates the lockfile from an `npm install` on Node 22 and NPM 11 (as is in the deploy environment) to unblock the installation.

This brings in the encoding package that the logs complain about not being depended upon.

No QA Required


chutchins25 (approved):
Looks fine to me.

PR #1281, opened 2026-02-23
chore: update to node 24

straker:
No QA required


.nvmrc
```diff
@@ -0,0 +1 @@
+24
```
straker, line 1: Decided to add this file since a lot of our other repos have it.

.github/workflows/deploy.yml
```diff
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           registry-url: 'https://registry.npmjs.org/'
-          node-version: 22
+          node-version: 24
```
Garbee, line 25:
```suggestion
          node-version-file: .nvmrc
```

Since you added the file, go ahead and leverage it in the workflows to reduce future update work.

packages/playwright/package.json
```diff
@@ -58,7 +58,7 @@
     "@types/chai": "^4.3.3",
     "@types/express": "^5.0.3",
     "@types/mocha": "^10.0.0",
-    "@types/node": "^22.0.2",
+    "@types/node": "^24.10.13",
```
Garbee, line 61: `@types/node` is not in the dependabot ignore list. It should be set to ignore majors. That way we get the updates we want without having noise from the ones we don't.

scottmries:
LGTM modulo other comments.

packages/playwright/package.json
straker: I'm not sure I agree. `@types/node` should be mostly backwards compatible, even across major versions, so accepting latest version should be fine even if we're not running our code on the latest version.

Garbee: Oh, this is not the devtools repo where we build the binaries. So this is less impactful.

Regardless, the types should be what we build and test on. Not "some potential future version" which will inevitably have new APIs as well.

"Mostly backwards compatible" is the key phrase, particularly the "mostly" part. To build reliable software we need to analyze and know what is on the platform we build to. We shouldn't rely on sheer dumb luck that things work out.

straker: The tests running successfully should be enough to know if the newest types are sufficient. It's only a dev dependency. We can pull this discussion into an issue rather than this pr if you'd like.

scottmries (approved):

PR #1288, opened 2026-03-03
chore: use nvmrc file for workflows

straker:
Moving to use the nvmrc file rather than hard code node version in the workflows.

No QA required

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           registry-url: 'https://registry.npmjs.org/'
-          node-version: 24
+          node-version-file: .nvmrc
           cache: 'npm'
       - name: Update npm to 11 if needed
         run: |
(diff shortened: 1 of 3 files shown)


package-lock.json
Garbee, line 1: Why is the lockfile changing at all with this?

straker: It shouldn't, but CI isn't working so I tried to see if updating the lockfile would fix the issue (it didn't).

> Run npm ci
> npm error code EUSAGE
> npm error
> npm error `npm ci` can only install packages when your package.json and package-lock.json or npm-shrinkwrap.json are in sync. Please update your lock file with `npm install` before continuing.
> npm error
> npm error Missing: encoding@0.1.13 from lock file

straker: I tried pulling in a fresh clone and doing `npm ci` and it worked locally, so I'm not sure why it's failing in ci

```bash
git clone https://github.com/dequelabs/axe-core-npm.git tmp-axe-core-npm
cd tmp-axe-core-npm
npm ci
```

Garbee: So, the [update deploy](https://github.com/dequelabs/axe-core-npm/actions/runs/22403948808/job/64858261074) from last week is also failing for the same reason. It started with the [json id update run](https://github.com/dequelabs/axe-core-npm/actions/runs/22229286256/job/64304204390).

Garbee: So yea update `npm` then install deps fixes that. But now unit tests are breaking. See: https://github.com/dequelabs/axe-core-npm/pull/1289

Garbee (approved):

PR #1280, opened 2026-02-16
chore: bump jsonld from 8.3.3 to 9.0.0

straker:
Manual upgrade as https://github.com/dequelabs/axe-core-npm/pull/1255 was stuck on couldn't resolve conflicts.

No QA required

PR #1289, opened 2026-03-04
chore: update lockfile

Garbee:
The lockfile got out of sync, in a very odd way. I think npm broke something in a minor in relation to how lockfiles get generated.

## The Problem

When [jsonLd](https://github.com/dequelabs/axe-core-npm/pull/1280) was updated, something about it [broke the builds](https://github.com/dequelabs/axe-core-npm/actions/runs/22229286256/job/64304204390). But only after getting up to the publish point. The [Node 24](https://github.com/dequelabs/axe-core-npm/pull/1281) upgrade got in fine, because it was probably slightly out of sync with `develop`.

## The Solution

Update to latest npm with `npm install -g npm` and then `npm install` to refresh the lockfile. Now encoding shows up. This 100% did not work before with an older NPM. So there is a break somewhere in expectations.

I narrowed it down to [npm 11.10.1](https://github.com/npm/cli/releases/tag/v11.10.1) that is where `encoding` started to get pulled in. Due to some "hosting" in that changelog, I'm wondering if this is a break from them moving dependencies around.

No QA Required

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -120,7 +120,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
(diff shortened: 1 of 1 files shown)


straker:
I'm going to pull this in just by itself so we can stop merging into separate prs to test things. Tests are failing for the chromedriver issue but this passes the linter.

straker (approved):

PR #1290, opened 2026-03-04
ci: debug chromedriver

straker:
Chromedriver is constantly failing with what I believe is the v146 update (both axe-core and axe-core-npm are now failing constantly to create chromedriver sessions). Adding verbose logging so I can see what's going on.

No QA required

PR #1291, opened 2026-03-04
ci: pin to chrome/driver 145

straker:
It appears chrome/driver v146 has issues and is consistently crashing or failing to create a session in our webdriverjs & cli tests (which uses webdriverjs). Trying to [figure out the problem didn't result in figuring out what went wrong](https://github.com/dequelabs/axe-core-npm/pull/1290) so I'm just pinning to v145 which still passes the tests. Hopefully this will be resolved in future versions, but I'll create a tech-debt ticket to look back at this.

This also brings in @Garbee changes [to fix the lockfile](https://github.com/dequelabs/axe-core-npm/pull/1289) as both problems are blocking tests.

No QA required

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
       - run: npm ci
       # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
       - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome
+        run: npx browser-driver-manager install chrome@145
         working-directory: packages/puppeteer
       - run: npm run build --workspace=packages/puppeteer
       - run: npm run coverage --workspace=packages/puppeteer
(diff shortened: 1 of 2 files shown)


Garbee:
I wonder if the 146 thing is because an older Chrome is on the runner images. Might be Chrome's control API had a breaking change and you need to update in lockstep.

One more reason we need to refactor all the CI eventually to use [browser-actions/setup-chrome](https://github.com/browser-actions/setup-chrome) where we don't use containers to run CI in. That action keeps the browser and driver in sync, regardless of what is on he host box.

Garbee (approved):

straker:
`browser-driver-manager` keeps things in lockstep by using `puppeteer/browser` and chrome for testing, so I'm not sure that would be the problem.

chutchins25 (approved):
Looks good to me! A nice, small change!

PR #1297, opened 2026-03-09
refactor: have all packages extend tsconfig from root

straker:
Continuing to clean up the repo a bit and having all `tsconfig` files extend from the root one. Also expanded the root `tsconfig` to have more strict options. Doing so revealed a type issue in cli when we tried to analyze the page and get the node style callback back so I changed the type to use a [control flow analysis for dependent parameters](https://devblogs.microsoft.com/typescript/announcing-typescript-4-6/#control-flow-analysis-for-dependent-parameters) which resolved the issue. This let me remove the type casts of the parameters and use what the function returned.

No QA required


Garbee:
From the test failures. Looks like https://github.com/dequelabs/axe-core-npm/blob/e1dd13a726423b26d7a4474447c8640bba3eb04d/packages/cli/src/bin/cli.ts#L4 needs to have `with { type: 'json' };` added to the import.

Then the playwright things seem to be failing due to code coverage. I have no idea why the changes here would impact the Playwright tests, unless they overlap with the CLI bit somehow. If the coverage data does show overlap for Playwright, we should address it here. Otherwise I'd lower the threshold slightly. I don't think the 2% is a big problem right now for that area.

straker:
Ya, what's even weirder is that odd changes to the root `tsconfig` breaks things completely, like adding `"noImplicitReturns": true,` makes some of the tests break. I haven't been able to find a reason that would be the case since no implicit returns doesn't affect tests as far as I'm aware

```
Exception during run: Error: Directory import '/home/runner/work/axe-core-npm/axe-core-npm/packages/webdriverio/src' is not supported resolving ES modules imported from /home/runner/work/axe-core-npm/axe-core-npm/packages/webdriverio/test/axe-webdriverio.spec.ts
```

Garbee:
Very interesting. I'll try to check things out in the morning based on wherever you end up tonight. I'm too tired after my day to have any more useful brain energy right now.

tsconfig.json
```diff
@@ -1,17 +1,24 @@
...
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
     "sourceMap": true,
-    "declaration": true,
-    "outDir": "dist",
-    "esModuleInterop": true,
-    "resolveJsonModule": true
+    "strict": true,
+    "strictNullChecks": true,
+    "target": "esnext"
```
chutchins25, line 20: Thank you for putting these in alphabetical order!

chutchins25 (dismissed):
Looks good to me.

packages/cli/package.json
```diff
@@ -80,10 +80,10 @@
       "text-summary",
       "html"
     ],
-    "statements": 95,
-    "branches": 91,
+    "statements": 90,
```
Garbee: I don't see the comments being removed in the current changeset. We should remove the instanbul ignore comments as well so we don't confuse anyone later.

straker: Decided to not remove comments for `ts-node` in order to keep the coverage as is and the comments. That way this pr stays small and focused and we could do that work in a different pr.

Garbee (approved):
LGTM from the last batch of changes and our discussion in slack. Thanks for getting a comment in from that before I could.

PR #1303, opened 2026-03-24
refactor(wdio): unlink browser type from global namespace

straker:
Deatches the Browser type from the global namespace and separates them by versions we support. This is a precursor to supporting v9 types which https://github.com/dequelabs/axe-core-npm/pull/1302 tried but failed to get the types correct. Am trying the same thing but without the conversion to v9 for now.

No QA required


Garbee (dismissed):
Of course the answer is an unknown cast first. Reasonable first step, let's see where the rest goes.

PR #1345, opened 2026-05-19
ci: replace browser-driver-manager with browser-actions/setup-chrome

Garbee:
## Summary

- Drops `browser-driver-manager` in favor of `browser-actions/setup-chrome` across every browser job in `tests.yml`. Default Chrome is `stable`, but the `CHROME_VERSION` GitHub repository variable can pin a version when needed.
- Standardizes Chrome / ChromeDriver discovery in every suite around two env vars: `CHROME_TEST_PATH` and `CHROMEDRIVER_TEST_PATH`. They are populated by the action and exported via `$GITHUB_ENV`.
- Fixes the failing `webdriverio` `test:export` job: `test/esmTest.mjs` now reads both env vars, so the spawned ChromeDriver and the launched Chrome come from the same `setup-chrome` install instead of mixing the pinned `chromedriver@^146` against whatever Chrome the runner image happens to ship (148 today).

### Per-suite changes
- **webdriverio**: removes the `loadBdmEnv` dotenv loader; `axe-webdriverio.spec.ts` and `test/esmTest.mjs` consume `CHROME_TEST_PATH` / `CHROMEDRIVER_TEST_PATH` directly.
- **webdriverjs**: drops the BDM cache-dir dotenv loader from `test-utils.ts`.
- **cli**: now reads `CHROME_TEST_PATH` / `CHROMEDRIVER_TEST_PATH` directly from `process.env`. The legacy BDM `~/.browser-driver-manager/.env` auto-load is preserved as a silent back-compat shim — see **Backwards compatibility** below. Existing `--chrome-path` / `--chromedriver-path` CLI flags and the `chromedriver` npm fallback are untouched.
- **puppeteer**: `puppeteerOpts()` and `test/esmTest.mjs` now set `executablePath` from `CHROME_TEST_PATH` when present.
- **playwright**: `chromium.launch()` conditionally takes `executablePath` from `CHROME_TEST_PATH`.
- **test/wdio**: `goog:chromeOptions.binary` is set from `CHROME_TEST_PATH` when present.

### Workflow changes (`.github/workflows/tests.yml`)
- Six jobs (`puppeteer`, `cli`, `webdriverjs`, `webdriverio`, `playwright`, `wdio_globals_test`) now run:
  ```yaml
  - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
    id: setup-chrome
    with:
      chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
      install-chromedriver: true
      install-dependencies: true
  - run: |
      echo "CHROME_TEST_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> "$GITHUB_ENV"
      echo "CHROMEDRIVER_TEST_PATH=${{ steps.setup-chrome.outputs.chromedriver-path }}" >> "$GITHUB_ENV"
  ```
- The Playwright job drops `npx playwright install --with-deps` since `install-dependencies: true` on `setup-chrome` covers the OS libs and Playwright launches our binary via `executablePath`.
- Root `browser-driver-manager` dep removed; `package-lock.json` refreshed.

### Public-facing documentation

Brings user-facing install guidance in line with the `@puppeteer/browsers` recommendation that test-assertion messages already use:

- `packages/cli/README.md`, `packages/webdriverjs/README.md`, `packages/webdriverio/README.md`, `packages/webdriverjs/error-handling.md`, `packages/webdriverio/error-handling.md`: BDM install instructions replaced with `npx @puppeteer/browsers install chrome@stable --path ~/.cache/puppeteer-browsers` (and the equivalent chromedriver command). The `--path` keeps binaries in a stable shared location instead of polluting whatever directory `npx` happens to be invoked from.
- Docs now also explain why callers must supply **both** the Chrome and ChromeDriver paths (either via `--chrome-path`/`--chromedriver-path` for the CLI or the equivalent setup-builder calls in the library packages): if only the new ChromeDriver is wired up, it auto-discovers the system Chrome — which almost never matches the new ChromeDriver version — and fails with `session not created`.
- `packages/cli/src/bin/index.ts:134-143`: the `SessionNotCreatedError` "ChromeDriver only supports" branch now recommends `npx @puppeteer/browsers install … --path ~/.cache/puppeteer-browsers` and instructs users to pass both `--chrome-path` and `--chromedriver-path`. Same trigger, same `process.exit(2)`; only the help text was updated.

### Backwards compatibility

Earlier versions of `@axe-core/cli` documented `browser-driver-manager` (BDM) as the install path and silently auto-loaded the env-var paths BDM wrote to `~/.browser-driver-manager/.env`. To avoid silently breaking existing users on that flow:

- BDM is no longer mentioned anywhere we surface — READMEs, runtime errors, error-handling docs, and test-setup assertions all point at `@puppeteer/browsers`.
- The CLI still auto-loads `~/.browser-driver-manager/.env` on startup if it exists (with `quiet: true` so it's a no-op when the file's absent). `dotenv` remains a CLI dependency for this shim only. `dotenv` doesn't override entries already in `process.env`, so shell-exported values and explicit flags always win.

Net effect: any setup that worked on `develop` keeps working on this branch. The other shipped packages (`@axe-core/{webdriverjs,webdriverio,puppeteer,playwright}`) only had test files and docs touched — no `src/` changes, no public API surface change.

No QA Required

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,7 +36,7 @@ jobs:
           node-version: ${{ matrix.node}}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
@@ -61,7 +61,7 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
@@ -84,7 +84,7 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
@@ -109,7 +109,7 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
@@ -169,7 +169,7 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
@@ -194,7 +194,7 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - uses: browser-actions/setup-chrome@v1
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
         id: setup-chrome
         with:
           chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,10 +36,15 @@ jobs:
           node-version: ${{ matrix.node}}
           cache: 'npm'
       - run: npm ci
-      # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
-      - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
-        working-directory: packages/puppeteer
+      - uses: browser-actions/setup-chrome@v1
+        id: setup-chrome
+        with:
+          chrome-version: ${{ vars.CHROME_VERSION || 'latest' }}
+          install-chromedriver: true
+          install-dependencies: true
+      - run: |
+          echo "CHROME_TEST_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> "$GITHUB_ENV"
+          echo "CHROMEDRIVER_TEST_PATH=${{ steps.setup-chrome.outputs.chromedriver-path }}" >> "$GITHUB_ENV"
       - run: npm run build --workspace=packages/puppeteer
       - run: npm run coverage --workspace=packages/puppeteer
       - run: npm run test:export --workspace=packages/puppeteer
@@ -56,10 +61,15 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-        # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
-      - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
-        working-directory: packages/cli
+      - uses: browser-actions/setup-chrome@v1
+        id: setup-chrome
+        with:
+          chrome-version: ${{ vars.CHROME_VERSION || 'latest' }}
+          install-chromedriver: true
+          install-dependencies: true
+      - run: |
+          echo "CHROME_TEST_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> "$GITHUB_ENV"
+          echo "CHROMEDRIVER_TEST_PATH=${{ steps.setup-chrome.outputs.chromedriver-path }}" >> "$GITHUB_ENV"
       - run: npm run coverage --workspace=packages/cli
 
   webdriverjs:
@@ -74,10 +84,15 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
-      - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
-        working-directory: packages/webdriverjs
+      - uses: browser-actions/setup-chrome@v1
+        id: setup-chrome
+        with:
+          chrome-version: ${{ vars.CHROME_VERSION || 'latest' }}
+          install-chromedriver: true
+          install-dependencies: true
+      - run: |
+          echo "CHROME_TEST_PATH=${{ steps.setup-chrome.outputs.chrome-path }}" >> "$GITHUB_ENV"
+          echo "CHROMEDRIVER_TEST_PATH=${{ steps.setup-chrome.outputs.chromedriver-path }}" >> "$GITHUB_ENV"
       - run: npm run build --workspace=packages/webdriverjs
       - run: npm run coverage --workspace=packages/webdriverjs
       - run: npm run test:export --workspace=packages/webdriverjs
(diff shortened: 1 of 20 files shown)


.github/workflows/tests.yml
```diff
@@ -36,10 +36,15 @@ jobs:
           node-version: ${{ matrix.node}}
           cache: 'npm'
       - run: npm ci
-      # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
-      - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
-        working-directory: packages/puppeteer
+      - uses: browser-actions/setup-chrome@v1
```
Garbee: The same setup-chrome block repeats in six jobs intentionally — composite actions aren't worth introducing for a five-line snippet, and YAML anchors don't play nicely with `uses:`. To override Chrome version repo-wide: set a `CHROME_VERSION` repository variable (Settings → Variables) to something like `120` or `beta`; unset it to track stable. `install-chromedriver: true` guarantees a version-matched driver beside the browser, and we hoist both outputs into `$GITHUB_ENV` so subsequent `npm run` steps inherit them — that's what fixes the original mismatch where the chromedriver npm pin and the runner-image Chrome had drifted apart.

.github/workflows/tests.yml
```diff
@@ -149,7 +169,15 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
-      - run: npx playwright install --with-deps
+      - uses: browser-actions/setup-chrome@v1
```
Garbee: Playwright now gets the same setup-chrome treatment instead of `npx playwright install --with-deps`. `install-dependencies: true` on the action covers the OS libs, and Playwright launches our binary via `executablePath` (see `axe-playwright.spec.ts`), so we never download Playwright's own Chromium build in CI. Local dev is unaffected because the spec falls back to Playwright's bundled Chromium when `CHROME_TEST_PATH` is empty.

.github/workflows/tests.yml
```diff
@@ -166,6 +194,15 @@ jobs:
           node-version: ${{ matrix.node }}
           cache: 'npm'
       - run: npm ci
+      - uses: browser-actions/setup-chrome@v1
```
Garbee: This job had no Chrome install step before. Now it gets the same env-var contract as the rest. Pair this with the `--no-sandbox` / `--disable-dev-shm-usage` flags added to `wdio.conf.ts` — without those the setup-chrome Chromium binary's zygote can't sandbox itself on Ubuntu 24 (restricted user-namespace clone) and Chrome exits immediately. That's the regression that initially failed this job after the workflow change.

packages/webdriverio/test/esmTest.mjs
```diff
@@ -8,7 +8,10 @@ import { fixturesPath } from 'axe-test-fixtures';
 import { spawn } from 'child_process';
 import { getFreePort, connectToChromeDriver } from './testUtils.js';
 
-const { default: { path: chromedriverPath } } = await import('chromedriver');
+const chromedriverPath =
```
Garbee, line 11: This is the actual fix for PR #1344. Previously this harness used the `chromedriver` npm package (pinned `^146`) and launched it against whatever Chrome the runner image had (148), producing the "session not created — only supports Chrome 146" failure. Reading `CHROMEDRIVER_TEST_PATH` first means CI uses the setup-chrome-installed driver (version-matched to `CHROME_TEST_PATH`); the npm package stays as the local-dev fallback so `npm test` without env vars still works.

packages/webdriverio/test/esmTest.mjs
```diff
@@ -35,7 +38,8 @@ async function integrationTest() {
       capabilities: {
         browserName: 'chrome',
         'goog:chromeOptions': {
-          args: ['--headless', '--no-sandbox']
+          args: ['--headless', '--no-sandbox'],
+          ...(chromeBinary && { binary: chromeBinary })
```
Garbee: Spread is conditional so the unset case leaves capabilities untouched and webdriverio's Selenium Manager finds a system Chrome. Setting `binary: undefined` here would override Selenium Manager and break local dev.

packages/puppeteer/test/utils.ts
```diff
@@ -44,5 +44,9 @@ export function puppeteerOpts(): LaunchOptions {
     options.args.push('--no-sandbox', '--disable-setuid-sandbox');
   }
 
+  if (process.env.CHROME_TEST_PATH) {
```
Garbee, line 47: Same fallback pattern as the playwright and wdio configs: when CI set `CHROME_TEST_PATH`, use that binary; otherwise let Puppeteer launch its bundled Chromium. The previous `/usr/bin/google-chrome-stable` hardcode in the sibling `esmTest.mjs` is gone for the same reason — system Chrome paths aren't portable across runner images, and now we don't need to guess.

packages/playwright/test/axe-playwright.spec.ts
```diff
@@ -49,7 +49,10 @@ describe('@axe-core/playwright', () => {
 
   beforeEach(async () => {
     browser = await chromium.launch({
-      args: ['--disable-dev-shm-usage']
+      args: ['--disable-dev-shm-usage'],
+      ...(process.env.CHROME_TEST_PATH && {
```
Garbee: Playwright supports launching arbitrary Chromium-compatible binaries via `executablePath`. Conditional spread preserves the local-dev path: unset env var → Playwright's bundled Chromium; set env var → setup-chrome's binary. No `--with-deps` install needed in the workflow because setup-chrome's `install-dependencies: true` handles the OS libs.

Garbee:
### Env-var contract for reviewers

Every browser test suite reads two env vars now:

- `CHROME_TEST_PATH` → passed to the launcher as the Chrome binary (`goog:chromeOptions.binary` for selenium/webdriverio, `executablePath` for puppeteer/playwright).
- `CHROMEDRIVER_TEST_PATH` → used by the suites that spawn chromedriver standalone (webdriverio export harness, webdriverjs, cli).

Both are populated by `browser-actions/setup-chrome` in CI and exported via `\$GITHUB_ENV`. Both are optional locally — unset means the library auto-discovers. Override the CI Chrome version by setting the `CHROME_VERSION` repository variable.

The original failure (PR #1344) was a Chrome/chromedriver version mismatch in `packages/webdriverio/test/esmTest.mjs`; the actual one-line fix is in that file (now reads `CHROMEDRIVER_TEST_PATH` first). Everything else in the diff is harmonizing the other suites + workflow so this doesn't recur the next time the runner image rolls.

test/wdio/wdio.conf.ts
```diff
@@ -62,7 +62,16 @@ export const config: Options.Testrunner = {
       // capabilities for local browser web tests
       browserName: 'chrome', // or "firefox", "microsoftedge", "safari"
       'goog:chromeOptions': {
-        args: ['--headless']
+        args: [
+          '--headless',
+          // Required for CI runners using >=Ubuntu 24.04
+          // @see https://github.com/SeleniumHQ/selenium/issues/14609
+          '--no-sandbox',
+          '--disable-dev-shm-usage'
```
Garbee, line 70: These flags didn't matter before this PR — wdio_globals_test had no explicit Chrome install and Selenium Manager auto-discovered the system Chrome with whatever defaults it needed. Now that we point wdio at setup-chrome's Chromium explicitly, Ubuntu 24's restricted user-namespace clone prevents the zygote from sandboxing itself and Chromium exits immediately on session create. Every other browser suite already passes `--no-sandbox` for the same reason (see linked Selenium issue) — this brings wdio in line.

.github/workflows/tests.yml
```diff
@@ -36,10 +36,15 @@ jobs:
...
       - run: npm ci
-      # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
-      - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
-        working-directory: packages/puppeteer
+      - uses: browser-actions/setup-chrome@2e1d749697dd1612b833dba4a722266286fbefcd # v2.1.2
+        id: setup-chrome
+        with:
+          chrome-version: ${{ vars.CHROME_VERSION || 'stable' }}
+          install-chromedriver: true
```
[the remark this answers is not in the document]
Garbee: Fixed the missed one from the previous update.

chutchins25 (commented):
Claude suggested several places that may be public facing and may need to be updated with install guidance.

- `packages/cli/README.md:17`
- `packages/webdriverjs/README.md:15`
- `packages/webdriverio/README.md:23`
- `packages/cli/src/bin/index.ts:134-140` (runtime error printed to axe-cli users when Chrome/ChromeDriver can't be located)

The new in-test assertion messages already point at `npx @puppeteer/browsers install …`. Would it be good to bring the above READMEs and CLI runtime message in line?

Garbee:
Doc updates are good catches. I'll look into it.

Runtime error though, not sure. I really need to look into the specifics of that one. The changes here shouldn't be too impactful to any existing code in this regard.

Drafting until I rectify those things.

chutchins25 (commented):
One last documentation update thought.

The runtime error in packages/cli/src/bin/index.ts:147 correctly tells users to pass both --chrome-path AND --chromedriver-path — because @puppeteer/browsers installs Chrome to ~/.cache/puppeteer (not on PATH), so the bundled chromedriver npm fallback can't auto-discover it. The README, however, only shows
  --chromedriver-path:
Each command prints the resolved binary path. Pass the ChromeDriver path to axe via --chromedriver-path <path>.

A user following the README verbatim would hit the exact SessionNotCreatedError that the runtime-message update is designed to recover from. Should we update the README so they agree for this case?

Garbee:
Looking into the readme thing. But `~/.cache/puppeteer` is incorrect. When you use this installer like this, it installs to the path you give it (or present working directory by default). So, that location isn't known in advance.

Let me try to get the messaging updated to be clear about that.

chutchins25 (approved):
Looks good to me now!

PR #1354, opened 2026-06-01
chore: increase timeout for axe-core update

Garbee:
It keeps timing out, give it more.

No QA Required


michael-siek (approved):

straker (approved):

Garbee:
Due to runner limitations going on right now, I am force merging this in. The change impacts no code under test in CI, so it is safe in all the checks we are held up on.

PR #1355, opened 2026-06-01
chore: pin node

Garbee:
Pin node to unblock installation due to a bug in 24.16.0 with some tools like playwright's dep for unzipping packages.

No QA Required


scottmries (approved):

Garbee:
Force merging since everything is broken and we need to get something done. This in will help unblock Steve with his fix for the browser setup.

PR #1356, opened 2026-06-01
test: fix webdriverio esmTest.mjs

straker:
`esmTest.mjs` was importing `chromedriver` from npm rather than relying on the globals. 

No QA required


Garbee (dismissed):
Test failure seems like a flake. Core changes LGTM for the webdriver area.

Garbee (approved):
LGTM

PR #1384, opened 2026-06-21
chore: bump Playwright to 1.61.0

gdarchen:
## Why

This updates `@axe-core/playwright` to the latest requested Playwright release so the package stays aligned with the current Playwright toolchain.

## What changed

- bump `@playwright/test` from `^1.60.0` to `^1.61.0` in `@axe-core/playwright`
- refresh the workspace lockfile entries for `@playwright/test`, `playwright`, and `playwright-core`

## Notes

- no source changes were needed beyond the dependency and lockfile refresh
- the package build and existing `@axe-core/playwright` test suite still pass with 1.61.0

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,6 +121,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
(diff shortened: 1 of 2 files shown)


Garbee:
We update dependencies through Dependabot normally. Also this is a development dependency on the maintainer side, it shouldn't have had any impact for consumers of the package. (I'm also doing a manual update of everything major after the pnpm update. Which, it seems updated this as a part of doing that in #1421.)

We appreciate the attempt to help keep things updated. In the future if any problems come up, please file an issue first to explain what situation you are having a problem with. That way we can try to understand the correct fix for the problem before code work is done.

Thank you.

PR #1421, opened 2026-08-23
chore: migrate from npm workspaces to pnpm

### LINKED ISSUE #1346: Migrate to pnpm 11
## Goal

Switch this repo's package manager to **pnpm 11**.

## Current state

- **Current lockfile:** `package-lock.json`
- **Workspaces monorepo:** yes
- **`packageManager` field:** not set

## Notes specific to this repo

This is a workspaces monorepo. Add `pnpm-workspace.yaml` and convert internal cross-package deps to `workspace:*`.

## Migration checklist

- [ ] Remove `package-lock.json`
- [ ] Run `pnpm import` to seed `pnpm-lock.yaml` from the prior lockfile's resolutions
- [ ] Add `"packageManager": "pnpm@11.x.x"` to root `package.json`
- [ ] Add `pnpm-workspace.yaml` and convert internal cross-package deps to `workspace:*`
- [ ] Commit `pnpm-lock.yaml`
- [ ] Update `.gitignore` if needed
- [ ] Update `.github/dependabot.yml` per the `api-team:dependabot-config` skill

## CI / deploy pipeline considerations

The following workflows exist in this repo and need a review pass; treat the list as a starting point — open the repo and confirm nothing was missed:

  - [ ] `.github/workflows/auto-patch-release.yml`
  - [ ] `.github/workflows/create-release-candidate.yml`
  - [ ] `.github/workflows/create-release.yml`
  - [ ] `.github/workflows/deploy.yml`
  - [ ] `.github/workflows/tests.yml`
  - [ ] `.github/workflows/update-axe-core.yml`
  - [ ] Each workflow's `actions/setup-node` block updated to `cache: 'pnpm'` (`pnpm-lock.yaml` must be committed in the same commit so the cache key resolves)
  - [ ] Every `npm ci` / `npm install` / `yarn install` replaced with `pnpm install --frozen-lockfile`
  - [ ] Release / publish workflows — registry publish remains `npm publish` (pnpm doesn't change that); verify any script wrapper that shells out
  - [ ] Downstream consumers of build artifacts — verify the output is byte-identical so anything depending on it keeps working

## Acceptance criteria

- [ ] Migration PR merges with CI green
- [ ] A release / deploy run (or canary) confirms the pipeline still ships
- [ ] `pnpm install --frozen-lockfile` succeeds locally on a clean checkout

## References

- `practices/dependencies.md`
- `plugins/api-team/skills/dependabot-config/SKILL.md`
- `plugins/api-team/skills/github-actions-workflow/SKILL.md`

Garbee:
npm workspaces hoists everything into one flat `node_modules`, so a package can import a dependency it never declared and nothing complains until publish. It also runs every install script. This moves the monorepo to pnpm 11: the isolated store means a package only sees what it declares, and `allowBuilds` makes install-time script execution an explicit allowlist.

The strictness paid for itself immediately — `sinon`, `dotenv`, `@types/sinon`, `@jest/globals`, and `@wdio/types` were all relying on hoisting and are now declared where they are used. The two cross-package dependencies moved to the workspace protocol. `package-lock.json` becomes `pnpm-lock.yaml`, `lerna.json` gains `npmClient: pnpm`, and the root pins `packageManager: pnpm@11.23.0`, which pnpm reads and self-switches on — so no Corepack, which Node 25 removes.

CI installs pnpm through a SHA-pinned `pnpm/action-setup` with `standalone: true`, so `@pnpm/exe`'s bundled Node satisfies pnpm's `engines.node: >=22.13` on the Node 20 leg. Installs are `--frozen-lockfile`, and every `--filter` carries `--fail-if-no-match` — bare `--filter` exits 0 on no match, which left the ten test jobs fail-open. `workspaceConcurrency: 2` stops five browser-driving packages launching at once. `minimumReleaseAge` adds a supply-chain cooldown just under Dependabot's seven days.

Node 18 is dropped; the matrix is 20/22/24. `chromedriver`'s `install.js` requires `proxy-agent`, ESM-only since v8.

`cheerio` is pinned to `1.0.0-rc.12`. 1.x is ESM, and the `browser` export condition jsdom selects cannot be required from enzyme's CJS. The pin lifts when enzyme is replaced.

Closes: #1346


.github/workflows/tests.yml
```diff
@@ -33,154 +38,181 @@ jobs:
...
         with:
           node-version: ${{ matrix.version || matrix.node }}
-          cache: 'npm'
-      - run: npm ci
+          cache: 'pnpm'
+      - run: pnpm install --frozen-lockfile
       # HACK: Force a TTY to enable browser-driver-manager to manipulate stdout.
       - shell: 'script -q -e -c "bash {0}"'
-        run: npx browser-driver-manager install chrome@145
+        run: pnpm exec browser-driver-manager install chrome@145
```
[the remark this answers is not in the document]
Garbee: Dismissed: verified empirically that this works. `pnpm exec` from a workspace subdirectory puts the workspace root's `node_modules/.bin` on PATH, so the root-declared `browser-driver-manager` resolves fine — running `pnpm exec browser-driver-manager --version` from `packages/cli` returns `2.0.1` with exit 0. No `-w` and no per-package dependency needed, and the `working-directory` is preserved as intended.

.github/scripts/prepare_release.sh
```diff
@@ -5,10 +5,18 @@ set -e
 
 releaseLevel="$1"
 
+# The calling actions provide no setup-node and no install step, so this script
+# has to provision pnpm itself. `lerna version` shells out to it once
+# `npmClient: pnpm` is set. @pnpm/exe bundles its own Node, so this works
+# regardless of the runner's Node version.
+if ! command -v pnpm > /dev/null; then
+  npm install -g @pnpm/exe@11
```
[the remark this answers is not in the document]
Garbee: Addressed in b57f4eb: pinned to `@pnpm/exe@11.23.0` (matching the root `packageManager`) and added `--ignore-scripts`.

.github/workflows/update-axe-core.yml
```diff
@@ -8,10 +8,17 @@ jobs:
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+      - uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
+        with:
+          standalone: true
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version-file: .nvmrc
       - uses: dequelabs/axe-api-team-public/.github/actions/create-update-axe-core-pull-request-v1@main
         with:
           token: ${{ secrets.PAT }}
```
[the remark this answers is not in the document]
Garbee: Addressed in b57f4eb: pinned to commit SHA `272a2eb86605251b73995fb158f30c3641e75593`. The four other `axe-api-team-public` actions still on `@main` (create-release, create-release-candidate, auto-patch-release, semantic-pr-footer) are pre-existing and out of scope for this PR.

package.json
```diff
@@ -2,18 +2,14 @@
...
+  "packageManager": "pnpm@11.23.0",
   "scripts": {
     "fmt": "prettier --write .",
     "lint": "eslint .",
     "prepare": "husky install",
     "precommit": "lint-staged",
-    "build": "npm run --workspaces build --if-present",
-    "test": "npm run --workspaces test --if-present"
+    "build": "pnpm -r run build",
+    "test": "pnpm -r run test"
```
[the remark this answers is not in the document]
Garbee: Dismissed: `pnpm -r run <script>` already behaves as `--if-present` — it runs the script only in packages that define it and does not error on the ones that don't. `pnpm -r run build` exits 0 locally with `test/wdio` and `test/axe-core` in the workspace, and CI is green.

packages/playwright/package.json
```diff
@@ -43,12 +43,12 @@
...
     "test": "mocha --timeout 60000 -r ts-node/register 'test/**.spec.ts'",
-    "test:export": "npm run test:esm && npm run test:commonjs && npm run test:ts",
+    "test:export": "pnpm run test:esm && pnpm run test:commonjs && pnpm run test:ts",
     "test:esm": "node test/esmTest.mjs",
     "test:commonjs": "node test/commonjsTest.js",
     "test:ts": "tsc test/tsTest.ts --noEmit --skipLibCheck --esModuleInterop",
-    "coverage": "nyc npm run test",
-    "prepare": "npx playwright install && npm run build"
+    "coverage": "nyc pnpm run test",
+    "prepare": "pnpm exec playwright install && pnpm run build"
```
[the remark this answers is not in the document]
Garbee: Dismissed: pnpm links the `playwright` bin into `packages/playwright/node_modules/.bin/` via the direct `@playwright/test` dependency, so `pnpm exec playwright` resolves. Verified locally (`pnpm exec playwright --version` → 1.60.0, exit 0) and in CI — the `playwright (20)`, `(22)`, and `(24)` jobs all pass on this commit.

.github/workflows/tests.yml
```diff
@@ -33,154 +38,181 @@ jobs:
...
           node-version: ${{ matrix.version || matrix.node }}
-          cache: 'npm'
-      - run: npm ci
-      - run: npx playwright install --with-deps
-      - run: npm run build --workspace=packages/playwright
-      - run: npm run coverage --workspace=packages/playwright
-      - run: npm run test:export --workspace=packages/playwright
+          cache: 'pnpm'
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm --filter @axe-core/playwright exec playwright install --with-deps
```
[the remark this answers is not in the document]
Garbee: Dismissed: same as the `packages/playwright/package.json` thread — the `playwright` bin is linked through the direct `@playwright/test` dependency and `pnpm --filter @axe-core/playwright exec playwright install` works. The `playwright (20)`, `(22)`, and `(24)` jobs pass on this commit, which exercises exactly this step.

.github/workflows/tests.yml
```diff
@@ -13,174 +13,206 @@ jobs:
...
+          standalone: true
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: ${{ matrix.version || matrix.node }}
-          cache: 'npm'
-      - run: npm ci
-      - run: npm run test --workspace=test/wdio
+          cache: 'pnpm'
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm --filter wdio-tests run test
```
straker: This is interesting and a bit confusing as we don't really care what the name of these packages are so I didn't know what this was trying to run. Too bad we can't filter by file path still.

.github/workflows/tests.yml
```diff
@@ -13,174 +13,206 @@ jobs:
...
+          standalone: true
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version-file: .nvmrc
-          cache: 'npm'
-      - run: npm ci
-      - run: npm test --workspace=test/axe-core
+          cache: 'pnpm'
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm --filter axe_core_test run test
```
straker: This one especially confused me as I thought it was trying to run the action step somehow. We should update the package name to not use underscores and be consistent there.

.github/workflows/tests.yml
```diff
@@ -13,174 +13,206 @@ jobs:
...
+          cache: 'pnpm'
+      - run: pnpm install --frozen-lockfile
+      - run: pnpm run lint
 
   puppeteer:
     strategy: &strategy_matrix
       fail-fast: false
       matrix:
-        node: [18, 20, 22, 24]
+        node: [20, 22, 24]
```
straker, line 33: The description said 18 was dropped because chromedriver moved to ESM, but that doesn't make any sense because `chromedriver` is only a binary when installed through `browser-driver-manager`. Additionally I can't find any information about that in the chromedriver repo and the [main lib file](https://github.com/giggio/node-chromedriver/blob/main/lib/chromedriver.js) and [install script](https://github.com/giggio/node-chromedriver/blob/main/install.js) are still using `require` (not that puppeteer uses the npm pacakge, but the CLI does and it would break if it was ESM).

straker: Looking at the action step something else is going on. [The action fails](https://github.com/dequelabs/axe-core-npm/actions/runs/32656289289/job/97235585760#step:5:35) in `.pnpm/chromedriver@152.0.1_debug@4.4.3_supports-color@8.1.1__supports-color@8.1.1/node_modules/chromedriver/install.js` when it tries to require `proxy-agent` and `proxy-agent` is ESM only. I'm not sure what that that package is or where it came from.

.github/workflows/tests.yml
Garbee: I mean, we can. It is just `working-directory: tests/wdio` then `run: pnpm run test`.

The syntax of keeping it directly through pnpm just changes it from a file path.

I think we just stick with the built-in `--filter` to keep everything consistent. As in the end everything has as package name. It's not a big issue. Most of the time these are running from scripts or by AI now anyways. Humans won't be typing things as often as we used to.

.github/workflows/tests.yml
Garbee: I think any package renames is separate work. To get a direct port done between package managers should be minimal.

I also have no clue how you'd think this is trying to run an action step. That's not how pnpm would work at all. I agree we should avoid underscores in package names as a consistency thing, but I don't think that would be in scope of this change set.

.github/workflows/tests.yml
Garbee: It is coming [directly from chromedriver as a dependency](https://npmx.dev/package-code/chromedriver/v/152.0.1/package.json#L37). And as of version 8 most likely, [proxy-agent is ESM only](https://npmx.dev/package-code/proxy-agent/v/8.0.2/package.json).

Thus, our dev dep is breaking running on Node 18. Unless we go force pin it back. Which is only going to cause issues.

We should just knock it off with these binary download packages. But, once again, conversion between package managers should be kept minimal. Doing anything about these binary downloaders needs to be separate.

Path of least resistance, just drop 18 in the matrix tests. It is EOL anyways.

.github/workflows/tests.yml
Garbee: ~~Oh, there is a `--workspace=<path>` option instead of `--filter`. I have it adjusting to that since it also fails closed instead of open.~~

This was incorrect. I got confused with the way Claude structured some wording. But it did add a `--fail-if-no-match` to close the fail-open gap.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,16 @@
+packages:
+  - 'packages/*'
+  - 'test/*'
+  - '!test/release'
+
+# Five packages drive real browsers in `test`, so bound how many run at once.
+workspaceConcurrency: 2
+
```
Zidious: We should add this -

```suggestion
# Supply-chain protection: refuse to install packages newer than ~6d23h,
# staying just under Dependabot's default 7-day cooldown.
# Integer in MINUTES. NOT a duration string like "10020m" (silently wrong).
minimumReleaseAge: 10020
```

Garbee: Oh right. Why did that get missed from the skill we have? I'll update my personal instructions while I'm doing these to make sure it gets in.

chutchins25 (approved):
Looks good to me.

straker (approved):

